PR #18, opened 2017-03-10
Chore/house keeping

axe312ger:
- [x] upgrade dependencies
- [x] switch to babel-preset-env
- [x] drop node support < v4.7
- [x] proper gitignore
- [x] replace ghooks with husky
- [x] replace standard cli with eslint with standard config
- [x] add trevor
- [x] clean up index.js and bin file


.babelrc
```diff
@@ -1,5 +1,19 @@
 {
-  "presets": ["es2015"],
+  "presets": [
+    ["env", {
+      "debug": true,
+      "targets": {
+        "node": 4.7,
+        "browsers": [
```
Khaledgarbaya: I am not sure if we should support browsers, content-import and export are mainly not intended to use in a browser env, I guess we could remove the `"browsers"` prop

package.json
```diff
@@ -1,19 +1,26 @@
...
   "name": "contentful-import",
+  "version": "0.0.0-placeholder",
   "description": "this tool allows you to import JSON dump exported by contentful-export",
-  "bin": "bin/contentful-import",
-  "main": "index.js",
-  "version": "0.0.1",
+  "bin": {
+    "contentful-import": "./bin/contentful-import"
+  },
+  "main": "dist/run-contentful-import.js",
```
Khaledgarbaya: is there a reason for the removal on `index.js`entirely ?

PR #24, opened 2017-03-21
enable the new rate limiting logic

axe312ger:


package.json
```diff
@@ -49,7 +49,7 @@
   },
   "dependencies": {
     "bluebird": "^3.3.3",
-    "contentful-batch-libs": "^4.9.2",
+    "contentful-batch-libs": "beta",
```
Khaledgarbaya: having the full version here is much better, once we remove the beta tag no changes will be required

lib/index.js
```diff
@@ -1,8 +1,8 @@
 import { find, filter } from 'lodash/collection'
-import createClients from 'contentful-batch-libs/utils/create-clients'
-import pushToSpace from 'contentful-batch-libs/push/push-to-space'
+import createClients from 'contentful-batch-libs/dist/utils/create-clients'
+import pushToSpace from 'contentful-batch-libs/dist/push/push-to-space'
```
Khaledgarbaya: not a big fan of this approach, but we can at a later time change this

Khaledgarbaya:
Just added a fix to a bug which was mentioned [here](https://github.com/contentful/contentful-import/issues/25)

PR #26, opened 2017-03-22
refactoring and cleanup

axe312ger:


README.md
```diff
@@ -72,9 +84,8 @@ spaceImport(options)
 
 ## Limitations
 
-- This tool imports only your published Content Model, Content and Assets. It will support Editor Interfaces, Webhooks and Roles & Permissions in a future version. Draft entries are not supported.
-- You can only import content to an empty space. It does not support merging content into pre-existent spaces.
-- If you have custom widgets, you need to reinstall them manually in the new space.
+- This tool currently does **not** support the import of roles & permissions and space memberships.
+- If you have custom UI extensions, you need to reinstall them manually in the new space.
```
Khaledgarbaya: Can you also mention that the webhooks with credential will be imported as a normal webhook and the credential should be entered manually

README.md
```diff
@@ -18,19 +18,31 @@ npm install -g contentful-import
...
+                                                             [string] [required]
+
+  --content-file             JSON file that contains data to be import to your
+                             space                           [string] [required]
+
+  --skip-content-model       Skips content types and locales. Copies only
+                             entries and assets                        [boolean]
+
+  --skip-locales             Skips locales. Must be used with
+                             content-model-only. Copies only content-types
```
madtrick: what about that `-` in `content-types`?

lib/usageParams.js
```diff
@@ -17,26 +17,16 @@ const opts = yargs
     demand: true
   })
   .option('content-file', {
-    describe: 'json file that contains data to be import to your space',
+    describe: 'JSON file that contains data to be import to your space',
     type: 'string',
     demand: true
   })
   .option('skip-content-model', {
-    describe: 'Skips content types and locales. Copies only entries and assets',
```
madtrick: Why did we remove this?

lib/usageParams.js
```diff
@@ -17,26 +17,16 @@ const opts = yargs
...
-    type: 'number',
-    default: 5
-  })
-  .option('rate-limit-period', {
-    describe: 'How much time to wait before retry in ms',
-    type: 'number',
-    default: 1000
-  })
   .option('skip-locales', {
-    describe: 'Skips locales. Must be used with content-model-only. Copies only content-types',
```
madtrick: Why did we remove this?

axe312ger: because there is no `content-model-only` flag (anymore?)

lib/usageParams.js
axe312ger: Cus it would still import other stuff like webhooks etc. So `only entries and assets` would be wrong.

madtrick: Good point

lib/index.js
```diff
@@ -6,16 +6,41 @@ import transformSpace from 'contentful-batch-libs/dist/transform/transform-space
...
+  if (!opts.spaceId) {
+    return Promise.reject(new Error('The `spaceId` option is required.'))
+  }
+
+  if (!opts.managementToken) {
+    return Promise.reject(new Error('The `managementToken` option is required.'))
+  }
+
+  if (!opts.contentFile && !opts.content) {
+    return Promise.reject(new Error('Eiter the `contentFile` or `content` option are required.'))
```
madtrick: Typo. Missing `h` -> `either`

madtrick:
LGTM

PR #33, opened 2017-04-07
Add info table about what to import and duration

axe312ger:
Besides adding the info table and a duration at the end, it tries to do implement a proper usage of `npmlog`

screen shot 2017-04-07 at 14 56 05
screen shot 2017-04-07 at 14 56 20

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -7,6 +7,11 @@ import getTransformedDestinationResponse from './get-transformed-destination-res
 import log from 'npmlog'
 import fs from 'fs'
 import { resolve } from 'path'
+import { startCase } from 'lodash'
+import Table from 'cli-table2'
+import moment from 'moment'
+
+const summary = {}
 
 export default function runContentfulImport (usageParams) {
   const defaultOpts = {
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -39,6 +46,28 @@ export default function runContentfulImport (usageParams) {
     return Promise.reject(new Error('Either the `contentFile` or `content` option are required.'))
   }
 
+  const infoTable = new Table()
+
+  infoTable.push([{colSpan: 2, content: 'The following amount of entities will be imported:'}])
```
Khaledgarbaya: I would show the summary after the import is done 
and instead of `The following amount of entities will be imported:` -> `the following entities were imported`

Khaledgarbaya:
@axe312ger can you also remove the `:` from the title

lib/index.js
axe312ger: I think it is more useful to see what is going to be imported before the actual import starts. So you get an overview what is going to happen.

We could display the same table at the end again, but it would display exactly the same since `pushToSpace` is not returning the entities that got successfully imported.

Khaledgarbaya:
@axe312ger I think showing what has been imported is better than showing what will be imported

PR #42, opened 2017-05-08
fix(parseOption): handle non existing lists

Khaledgarbaya:
When the json is changed manually and one of the arrays that the tool is expecting is removed the tool will crash, this PR fixes that

### DIFF
--- a/lib/parseOptions.js
+++ b/lib/parseOptions.js
@@ -50,8 +50,9 @@ export default function parseOptions (params) {
   options.destinationManagementToken = options.managementToken
   options.content = options.content || JSON.parse(fs.readFileSync(options.contentFile))
 
-  // to avoid loping through an undefined object later on
+  // to avoid looping through an undefined array later on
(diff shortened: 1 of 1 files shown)


lib/parseOptions.js
```diff
@@ -49,7 +49,14 @@ export default function parseOptions (params) {
   options.destinationSpace = options.spaceId
   options.destinationManagementToken = options.managementToken
   options.content = options.content || JSON.parse(fs.readFileSync(options.contentFile))
+
+  // to avoid loping through an undefined object later on
```
axe312ger: little type, even when it's just in the comment 🙈

lib/parseOptions.js
```diff
@@ -49,7 +49,14 @@ export default function parseOptions (params) {
...
+  options.content.contentTypes = options.content.contentTypes || []
```
axe312ger: I am missing `entries` here

axe312ger:
squash and 🚢  it!

PR #48, opened 2017-06-01
[WIP] Refactor info and error logging

axe312ger:
* errors are logged to error log file
* single errors do not stop the import anymore (add a flag to reestablish behavior? but when should it break then?)
* ~now aborts infinite publishing queues~ (fixed in other PR)

Hint: This PR is yet WIP and follows to principle to do as least changes as possible to achieve the goal. Further rewrites might be needed.

#### Current state logging:
screen shot 2017-06-01 at 17 26 57

#### Current state error log file content:
screen shot 2017-06-01 at 17 26 47


axe312ger:
I think rewriting https://github.com/contentful/contentful-batch-libs/blob/master/lib/push/push-to-space.js to use listr would be neccessary, also.

lib/index.js
```diff
@@ -1,98 +1,152 @@
...
 export default function runContentfulImport (params) {
-  summary.startTime = moment()
+  const summary = {
+    startTime: moment()
+  }
+  const errorLog = []
+  const timestamp = summary.startTime.format('YYYY-MM-DDTHH-mm-SS')
 
   const options = parseOptions(params)
+  const errorLogFilePath = `${process.cwd()}/contentful-import-error-log-${options.spaceId}-${timestamp}.json`
```
Khaledgarbaya: we have `--error-log-file`, we should check if it is defined otherwise we define our own, maybe move all of this to `parseOptions` module

lib/parseOptions.js
```diff
@@ -47,6 +50,12 @@ export default function parseOptions (params) {
     throw new Error('Please provide the proxy config in the following format:\nhost:port or user:password@host:port')
   }
 
+  options.startTime = moment()
+
+  if (!options.errorLogFile) {
```
Khaledgarbaya: you need ` resolve(process.cwd(), options.errorLogFile)` also if it is set, because user may pass in `mylogfile.json` fro example

axe312ger: this is done btw ^^

axe312ger:
closing in favor of #58

PR #53, opened 2017-06-15
fix(contentful-import): fix proxy

Khaledgarbaya:


lib/utils/proxy.js
```diff
@@ -0,0 +1,57 @@
...
+    port: portString,
+    auth: authString,
+    protocol
+  } = parse(proxyString)
+
+  const auth = parseAuth(authString)
+  const port = toInteger(portString)
+
+  if (!auth.username) {
+    return { host, port }
```
zcei: `isHttps: protocol === 'https:'` should be in here as well

PR #58, opened 2017-07-11
[next] proper error logging

axe312ger:
Based on https://github.com/contentful/contentful-batch-libs/pull/126 which is now released via the `next` tag to npm as v6

* Implements new features of https://github.com/contentful/contentful-batch-libs/pull/126
* upgrades dependencies
* improves tests
* moves result table to the top since it actually represents the data to import, not the data imported. Another table at the end with the imported data might be useful, we may add it later on.
* introduces listr to the import tool


lib/index.js
```diff
@@ -1,98 +1,159 @@
...
-        skipContentPublishing: options.skipContentPublishing
-      })
-    })
-    .then((response) => {
-      log.info('import', 'Finished importing all data')
+  // Setup custom log listener to store log messages for later
+  setupLogging(log)
 
-      const infoTable = new Table()
+  console.log('This is a Beta release')
```
Khaledgarbaya: Same as the export tool, you need to add that `epilog`

PR #65, opened 2017-10-19
Next major version

axe312ger:
Features
* Proper error logfile which should give now way more context when imports go wrong.
* Beautiful and useful output via [listr](https://github.com/SamVerschueren/listr). With automatic CI detection.
* Overview table of what was imported
* Several bugfixes

Not done yet, more to come. Available via `npm install contentful-import@next`. Stress testing, suggestions, ideas, criticism very welcome :)


axe312ger:
We need to fix the following dependencies:
```
Unused dependencies
* http-proxy-agent
* json-stringify-safe
Unused devDependencies
* babel-eslint
* coveralls
* eslint-plugin-import
* eslint-plugin-node
* husky
* require-all
Missing dependencies
* listr
* listr-update-renderer
* listr-verbose-renderer
* tape
```

lib/index.js
```diff
@@ -43,6 +43,10 @@ export default function runContentfulImport (params) {
   infoTable.push([{colSpan: 2, content: 'The following entities are going to be imported:'}])
 
   Object.keys(options.content).forEach((type) => {
+    if (type === 'roles') {
```
Khaledgarbaya: I would define a list of what the tool can import and filter through that in the `parseOptions`.

PR #67, opened 2017-11-22
fix: Handle content file fs errors

Khaledgarbaya:


lib/parseOptions.js
```diff
@@ -63,7 +63,11 @@ export default function parseOptions (params) {
   // Further processing
   options.destinationSpace = options.spaceId
   options.destinationManagementToken = options.managementToken
-  options.content = options.content || JSON.parse(fs.readFileSync(options.contentFile))
+  try {
+    options.content = options.content || JSON.parse(fs.readFileSync(options.contentFile))
+  } catch (e) {
+    throw new Error(`Unable to access the content file, Make sure the file exist and you have the right permission for it`)
```
axe312ger: unnecessary template string

axe312ger: also `Make` should be `make`

axe312ger:
@Khaledgarbaya can you please rebase this one? :)

PR #78, opened 2018-01-23
feat: Add default locale check

Khaledgarbaya:
This PR adds an assert for default locales between destination space and source space.
the import tool will stop importing and tell you to change the default locale if they don't match.

### DIFF
--- a/lib/utils/validations.js
+++ b/lib/utils/validations.js
@@ -1,0 +1,20 @@
+const assertDefaultLocale = (source, destination) => {
+  const sourceDefaultLocale = source.locales.find((locale) => locale.default === true)
+  const destinationDefaultLocale = destination.locales.find((locale) => locale.default === true)
+
+  if (!sourceDefaultLocale || !destinationDefaultLocale) {
+    return
+  }
+
+  if (sourceDefaultLocale.code !== destinationDefaultLocale.code) {
+    throw new Error(`
+      Please make sure you destination space have the same default locale as the source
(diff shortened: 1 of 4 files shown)


lib/utils/validations.js
```diff
@@ -0,0 +1,20 @@
...
+  const sourceDefaultLocale = source.locales.find((locale) => locale.default === true)
+  const destinationDefaultLocale = destination.locales.find((locale) => locale.default === true)
+
+  if (!sourceDefaultLocale || !destinationDefaultLocale) {
+    return
+  }
+
+  if (sourceDefaultLocale.code !== destinationDefaultLocale.code) {
+    throw new Error(`
+      Please make sure you destination space have the same default locale as the source
```
axe312ger: `you` -> `your`

axe312ger: Maybe add newline at the end?

lib/utils/validations.js
```diff
@@ -0,0 +1,20 @@
...
+      default locale for source space : ${sourceDefaultLocale.code}\n
```
axe312ger: `Default` and extra space in front of the colon character

lib/utils/validations.js
```diff
@@ -0,0 +1,20 @@
...
+      default locale for destination space: ${destinationDefaultLocale.code}
```
axe312ger: `Default`

axe312ger:
LGTM. 🚢  it after CI turns green again

PR #81, opened 2018-01-25
WIP feat: paylod validation

Khaledgarbaya:
- [x] Add more tests
- [x] Improve error message?


Khaledgarbaya:
aaaaaand `Joi` does not support node `v4.7` 🤦‍♂️

lib/utils/schema.js
```diff
@@ -0,0 +1,76 @@
...
+  fields: Joi.object()
+}
+
+const contentTypeSchema = {
+  sys: Joi.object(),
+  fields: Joi.array().required().items(Joi.object().keys({
+    id: Joi.string().required(),
+    name: Joi.string().required(),
+    type: Joi.string().required().regex(/Symbol|Text|Integer|Number|Date|Object|Boolean|Array|Link/),
+    validations: Joi.array(),
```
axe312ger: Shouldn't we have a schema for validations?

Khaledgarbaya: validations is so vague and it depends on the field. We can live without validating the validations for now, also we didn't have any issue related to the validations

axe312ger: Alright, sounds good to me

Khaledgarbaya:
@axe312ger this PR is ready for review, `Joi` by default was stopping early, that's why it only shows one error. I disabled that and the error message telling you how many entities are invalid.

axe312ger:
Neat, seeing all errors at once is amazing 🎉 

But the output in the CLI still could use some readability:
```
15:05:85 - ValidationError: Payload validation error: Invalid assets:7 Invalid locales:1 Invalid webhooks:1 Get further details in the error log file
```

How about:
```
15:05:85 - ValidationError: 7 assets, 1 locales, 1 webhooks - Get further details in the error log file
```
(Maybe even check if it is multiple and append `s` to make it plural)

lib/utils/validations.js
```diff
@@ -19,17 +19,15 @@ const countInvalidEntities = (validationData) => {
...
   }, {})
-  return Object.keys(entityCount).map((key) => `Invalid ${key}:${entityCount[key]}`)
+  return Object.keys(entityCount).map((key) => `${key}:${entityCount[key]}`)
 }
 
 const assertPayload = (payload) => {
   const result = Joi.validate(payload, payloadSchema, {allowUnknown: true, abortEarly: false})
   if (result.error) {
     attachEntityName(result.error.details, payload)
     const invalidEntityCount = countInvalidEntities(result.error.details).join('\n')
```
axe312ger: I'd join by `, ` not by newline. We have only one line available here.

errorlogfile
```diff
@@ -0,0 +1,106 @@
+[
```
axe312ger: Is this file needed? I don't think so 🤷‍♂️

lib/utils/validations.js
```diff
@@ -1,3 +1,38 @@
...
+    return entities
+  }, {})
+  return Object.keys(entityCount).map((key) => `${key}:${entityCount[key]}`)
+}
+
+const assertPayload = (payload) => {
+  const result = Joi.validate(payload, payloadSchema, {allowUnknown: true, abortEarly: false})
+  if (result.error) {
+    attachEntityName(result.error.details, payload)
+    const invalidEntityCount = countInvalidEntities(result.error.details).join('\n')
```
axe312ger: Lets join via comma + space instead of new line since we have only one line available in the Listr task error display.

test/unit/validations.test.js
```diff
@@ -0,0 +1,181 @@
...
+      contentType: 'image/png'
+    }
+  }
+  expect(() => assertPayload({
+    entries: [],
+    locales: [],
+    contentTypes: [],
+    assets: [brokenAsset, brokenAsset]
+  })
+  ).toThrowErrorMatchingSnapshot()
```
axe312ger: Jest ❤️

lib/utils/validations.js
```diff
@@ -1,3 +1,38 @@
...
+const countInvalidEntities = (validationData) => {
+  const entityCount = validationData.reduce((entities, currentDetail) => {
+    if (!entities[currentDetail.path[0]]) {
+      entities[currentDetail.path[0]] = 1
+    } else {
+      entities[currentDetail.path[0]]++
+    }
+    return entities
+  }, {})
+  return Object.keys(entityCount).map((key) => `${key}:${entityCount[key]}`)
```
axe312ger: What about if we pass key through https://lodash.com/docs/4.17.4#startCase. IMHO that would improve readability :)

(We probably could do this then later on on more parts in the code to properly format entity types)

test/unit/index.test.js
```diff
@@ -5,10 +5,15 @@ import pushToSpaceStub from 'contentful-batch-libs/dist/push/push-to-space'
...
 import createClientsStub from 'contentful-batch-libs/dist/utils/create-clients'
 import getDestinationResponseStub from '../../lib/get-destination-response'
-
 import contentfulImport from '../../lib/index'
 
 jest.mock('cli-table2')
+jest.mock('../../lib/utils/validations', () => {
+  return {
+    assertPayload: jest.fn(),
+    assertDefaultLocale: jest.fn().mockImplementationOnce(() => { throw new Error('ContentfulMltiError') })
```
axe312ger: `ContentfulMultiError` - u missed the u

lib/parseOptions.js
```diff
@@ -57,7 +57,7 @@ export default function parseOptions (params) {
 
   const proxySimpleExp = /.+:\d+/
   const proxyAuthExp = /.+:.+@.+:\d+/
-  if (options.proxy && !(proxySimpleExp.test(options.proxy) || proxyAuthExp.test(options.proxy))) {
+  if (typeof options.proxy === 'string' && options.proxy && !(proxySimpleExp.test(options.proxy) || proxyAuthExp.test(options.proxy))) {
```
axe312ger: Can you duplicate this to https://github.com/contentful/contentful-export/blob/master/lib/parseOptions.js#L44 plz? kthxbye

lib/utils/validations.js
Khaledgarbaya: my idea is to map this key to the key in the JSON payload. I think it is fine like that

axe312ger:
:shipit:

PR #94, opened 2018-03-05
Add Environment support

Khaledgarbaya:
This PR adds support for the environment API.

- [x] Check tests

### DIFF
--- a/lib/tasks/push-to-space/push-to-space.js
+++ b/lib/tasks/push-to-space/push-to-space.js
@@ -85,7 +85,7 @@ export default function pushToSpace ({
       title: 'Importing Locales',
       task: wrapTask((ctx, task) => {
         return creation.createEntities(
-          { space: ctx.environment, type: 'Locale' },
+          { target: ctx.environment, type: 'Locale' },
           sourceData.locales,
           destinationData.locales
         )
--- a/lib/tasks/push-to-space/push-to-space.js
+++ b/lib/tasks/push-to-space/push-to-space.js
@@ -99,7 +99,7 @@ export default function pushToSpace ({
       title: 'Importing Content Types',
       task: wrapTask((ctx, task) => {
         return creation.createEntities(
-          {space: ctx.environment, type: 'ContentType'},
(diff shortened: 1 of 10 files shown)


lib/tasks/push-to-space/push-to-space.js
```diff
@@ -238,7 +240,7 @@ export default function pushToSpace ({
       title: 'Creating Web Hooks',
       task: wrapTask((ctx, task) => {
         return creation.createEntities(
-          {space: ctx.space, type: 'Webhook'},
+          {environment: ctx.environment, type: 'Webhook'},
```
axe312ger: this should be still space since web hooks life within spaces

lib/tasks/push-to-space/push-to-space.js
```diff
@@ -199,7 +202,7 @@ export default function pushToSpace ({
       title: 'Importing Content Entries',
       task: wrapTask((ctx, task) => {
         return creation.createEntries(
-          {environment: ctx.environment, skipContentModel},
+          {space: ctx.environment, skipContentModel},
```
axe312ger: I did call the space property in export `source` since it can bei either a space or an environment.

I suggest to use `target` here :)

Khaledgarbaya: I was avoiding editing all the other files, but yeah I guess I have to do it 😅

axe312ger: 🚀 it! :D

Khaledgarbaya:
Tests should be green now

lib/tasks/push-to-space/push-to-space.js
```diff
@@ -73,14 +74,18 @@ export default function pushToSpace ({
...
+            ctx.environment = environment
           })
       })
     },
     {
       title: 'Importing Locales',
       task: wrapTask((ctx, task) => {
         return creation.createEntities(
-          { space: ctx.space, type: 'Locale' },
+          { space: ctx.environment, type: 'Locale' },
```
axe312ger: should be target

PR #115, opened 2018-10-11
fix(webhooks): exclude secret headers

phoebeschmidt:
For security reasons, the value of secret headers are not exported with webhooks. This was causing an invalid webhook body on the import (and a 500 error). The fix is to exclude secret headers from the import.

### DIFF
--- a/lib/transform/transformers.js
+++ b/lib/transform/transformers.js
@@ -20,6 +20,16 @@ export function webhooks (webhook) {
   if (webhook.httpBasicUsername) {
     delete webhook.httpBasicUsername
   }
+
+  // Workaround for webhooks with secret headers
+  if (webhook.headers) {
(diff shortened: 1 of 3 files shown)


lib/transform/transformers.js
```diff
@@ -20,6 +20,16 @@ export function webhooks (webhook) {
   if (webhook.httpBasicUsername) {
     delete webhook.httpBasicUsername
   }
+
+  // Workaround for webhooks with secret headers
+  if (webhook.headers) {
+    webhook.headers = webhook.headers.reduce((accumulator, header) => {
```
Khaledgarbaya: Any reason to not use `Array.prototype.filter` instead?

```
webhook.headers = webhook.headers.filter((header) => !header.secret)
```

phoebeschmidt: nope just slipped my mind :) will change

PR #120, opened 2019-04-16
sort locales by fallback code

mariobodemann:


lib/utils/sort-locales.js
```diff
@@ -0,0 +1,36 @@
...
+      localeByFallback[locale.fallbackCode] = []
+    }
+
+    localeByFallback[locale.fallbackCode].push(locale)
+  })
+
+  return sortByFallbackKey(undefined, localeByFallback)
+}
+
+function sortByFallbackKey (key, localeByFallback) {
```
makinwab: Given the parameter `key` can be undefined, `sortByFallbackKey` should probably have the `key` parameter specified as the last parameter.

I.e. 
```
function sortByFallbackKey(localeByFallback, key)
```

and invocation on line 20, would look like this
```
sortByFallbackKey(localeByFallback)
```

`key` would still be undefined.

lib/utils/sort-locales.js
```diff
@@ -0,0 +1,36 @@
...
+  if (!localeByFallback[key]) {
+    return []
+  } else {
```
makinwab: Maybe a nitpick, there's probably no need for the `else` branch since you'll be jumping out of the function if the condition above is true.

test/unit/sort-locales.test.js
```diff
@@ -0,0 +1,99 @@
...
+    fallbackCode: 'b'
+  },
+  {
+    code: 'c',
+    fallbackCode: 'base'
+  }
+]
+
+test('Sorts locales by "fallback" order', () => {
+  const sortedEntries = sortLocales(locales)
```
makinwab: Did you mean `sortedLocales` instead of `sortedEntries`?

test/unit/sort-locales.test.js
```diff
@@ -0,0 +1,99 @@
...
+    fallbackCode: 'b'
+  },
+  {
+    code: 'd',
+    fallbackCode: 'a'
+  }
+]
+
+test('Exclude circular references', () => {
+  const sortedEntries = sortLocales(circle)
```
makinwab: Same as the comment above.

lib/utils/sort-locales.js
```diff
@@ -0,0 +1,36 @@
...
+    const resolved = localeByFallback[key]
```
makinwab: Does `resolved` hold sorted locales in the long run? If yes, can we name this `sortedLocales` or something else that's more explicit maybe?

lib/utils/sort-locales.js
mariobodemann: Both work the same for me. So I removed the else, it has a more guarded style now, and yes, it feels closer aligned with the rest of the code base. :+1:

test/unit/sort-locales.test.js
mariobodemann: Yeah, that looks like a copy'n'past error, thanks for spotting.

lib/utils/sort-locales.js
mariobodemann: Sounds good.

PR #220, opened 2020-05-27
Expand timeout and retryLimit params to asset handling

ruderngespra:


lib/parseOptions.js
```diff
@@ -16,6 +16,7 @@ const SUPPORTED_ENTITY_TYPES = [
 ]
 
 export default function parseOptions (params) {
+  // TODO: Do we need defaults here for timeout and retryLimit?
```
phoebeschmidt, line 19: I think no. We should let contentful-management sdk provide its own defaults, otherwise we'll always override them with whatever defaults we set here, which could be confusing in the future

lib/tasks/push-to-space/push-to-space.js
```diff
@@ -163,7 +165,8 @@ export default function pushToSpace ({
         )
           .then(removeEmptyEntities)
           .then((assetsToProcess) => {
-            return assets.processAssets(assetsToProcess)
+            // TODO: What are the default values of these parameters? Do we need defaults?
```
phoebeschmidt, line 168: hm I'd prefer to not send them if they aren't present in `option`s object. So if the user doens't provide them, we don't send them. (and let the cma provide it's own defaults, like I mentioned above).

There's isn't really a pretty way to do this, though. Maybe you can come up with something better:

```
const processingOptions = Object.assign({}, timeout && {timeout}, retryLimit && {retryLimit})
return assets.processAssets(assetsToProcess, processingOptions)...
```

test/unit/tasks/push/push-to-space.test.js
```diff
@@ -123,6 +125,8 @@ test('Push content to destination space', () => {
       expect(publishing.archiveEntities.mock.calls).toHaveLength(2)
       expect(editorInterfaceUpdateMock.mock.calls).toHaveLength(1)
       expect(assets.processAssets.mock.calls).toHaveLength(1)
+      expect(assets.processAssets.mock.calls).toHaveLength(1)
+      expect(assets.processAssets.mock.calls[0][1]).toEqual({retryLimit: 20, timeout: 40000})
```
phoebeschmidt, line 129: perfect

lib/tasks/push-to-space/push-to-space.js
ruderngespra: Thanks, makes sense! I will take this for now :)

PR #416, opened 2021-03-17
Import exported asset files

damienxy:
- Add new flags: `uploadAssets`, `assetsDirectory`, `rateLimit`
- Add functionality to upload assets from local files exported via `downloadAssets` option of contentful-export.
- Refactor promises to async await
- Use `pqueue` for rate limiting
- Bump node version to 12


marcolink (commented):
This looks really great 🎉 

Should we add some more tests around asset imports?

PR #467, opened 2021-05-25
fix: preserve tag visibility

TimBeyer:
Ensure that tag visibility is preserved

### DIFF
--- a/test/unit/tasks/push/creation.test.js
+++ b/test/unit/tasks/push/creation.test.js
@@ -188,6 +188,115 @@ test('Create entries and handle regular errors', () => {
     })
 })
 
+test('Create private tags', () => {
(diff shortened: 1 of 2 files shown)


test/unit/tasks/push/creation.test.js
```diff
@@ -188,6 +188,115 @@ test('Create entries and handle regular errors', () => {
     })
 })
 
+test('Create private tags', () => {
```
makinwab, line 191: Can the visibility value in the original be different from the visibility in the transformed data?

TimBeyer: No, the transformer just returns `tag` as it came in

makinwab (approved):

PR #591, opened 2021-12-01
feature(headers): Add standard header and fix tests

andreascful:
Add standard header and fix tests


.github/dependabot.yml
```diff
@@ -1,17 +1,48 @@
...
+  ignore:
+  - dependency-name: "@types/node"
+    versions:
+    - ">=17.0.0"
+  commit-message:
+    prefix: build
+    include: scope
+  registries:
+  - npm-registry-registry-npmjs-org
+- package-ecosystem: docker
```
marcolink, line 25: I think we can safely remove th docker part here

test/integration/import-lib.test.js
```diff
@@ -16,71 +16,95 @@ jest.setTimeout(1.5 * 60 * 1000) // 1.5min timeout
 
 beforeEach(async () => {
   const client = createClient({ accessToken: managementToken })
-  space = await client.createSpace({ name: 'IMPORT [AUTO] TOOL' }, orgId)
+  space = await client.createSpace({ name: 'IMPORT [AUTO] TOOL TMP' }, orgId)
```
marcolink, line 19: Why is this change relevant?

andreascful: Since the internal testing setup is still chaotic in parts (many different spaces in different testing orgs) it sometimes happens that e.g. spaces are not properly deleted after a test. I thought this might be an idicator for someone looking through the spaces to see that this is just a temporary leftover space. But I don't insist on this change.

PR #602, opened 2022-06-02
chore(deps): replace moment with date-fns

Brad-Turner:
Replaces moment with date-fns. This is to match the work done in the contentful/export repository (see https://github.com/contentful/contentful-export/pull/968) for more details.


marcolink (approved):
Thanks @Brad-Turner 🎉

PR #603, opened 2022-06-02
fix(test): modify unit tests to handle x-platform regex matching

Brad-Turner:
PR is to adjust regex path separators into a more cross platform solution. This copies the work done on the contentful-export project: https://github.com/contentful/contentful-export/pull/945/files


marcolink:
@Brad-Turner 🎉

PR #604, opened 2022-06-02
chore(build): add babel preset-env

Brad-Turner:
Updates the babel config with the same configuration as that found in the contentful-export tool.

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,10 +1,13 @@
 {
   "presets": [
-    ["env", {
-      "targets": {
-        "node": "14"
+    [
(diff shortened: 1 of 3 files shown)


babel.config.json
```diff
@@ -1,10 +1,13 @@
 {
   "presets": [
-    ["env", {
-      "targets": {
-        "node": "14"
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "node": "14"
```
Brad-Turner, line 7: I left this as node:14, its worth noting the package json supports node12+

PR #642, opened 2022-10-06
feat: use vault secrets [SDK-2258]

crissto:
Made everything one PR because the only change is when releasing. So no need for the two steps

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,8 @@
 version: 2.1
+
+orbs:
+  vault: contentful/vault@1
+
 jobs:
   unit:
     docker:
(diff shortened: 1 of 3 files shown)


andreascful:
I guess the reason why it is failing at the moment is because the vault server has not yet picked up your vault config, therefore it is not able to execute the get-secrets step, since the authentication is failing.

crissto:
> I guess the reason why it is failing at the moment is because the vault server has not yet picked up your vault config, therefore it is not able to execute the get-secrets step, since the authentication is failing. You might want to put the ci config in a separate PR.

Yeahh I tried to be fancy but didn't realise that it releases on PR. Did https://github.com/contentful/contentful-import/pull/643 for the initial setup

andreascful (approved):
👍

PR #715, opened 2023-01-20
fix: parsing of large files > 512 MB [ZEND-3117]

marcolink:
Parsing large input files (> 512 MB) caused `JSON.parse` to throw an error:
```
Error: Cannot create a string longer than 0x1fffffe8 characters
```
this is a node internal hard limit for parsing strings to objects. 

This PR introduces a streaming library to allow bigger input files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -197,5 +197,3 @@ $RECYCLE.BIN/
 # ignore left over error files
 errorlogfile.json
 
-# test input
-large-space.json
--- a/.gitignore
+++ b/.gitignore
@@ -196,3 +196,6 @@ $RECYCLE.BIN/
 
 # ignore left over error files
 errorlogfile.json
+
+# test input
+large-space.json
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -196,3 +196,6 @@ $RECYCLE.BIN/
 
 # ignore left over error files
 errorlogfile.json
+
+# test input
+large-space.json
```
andreascful, line 201: Should we really have this permanently in the gitignore? :)

lib/parseOptions.js
```diff
@@ -74,10 +75,12 @@ export default function parseOptions (params) {
   } else {
     options.errorLogFile = resolve(process.cwd(), options.errorLogFile)
   }
-
-  // Further processing
   options.accessToken = options.managementToken
-  options.content = options.content || JSON.parse(fs.readFileSync(options.contentFile))
+
+  if (!options.content) {
```
andreascful, line 80: Should we maybe add a comment here briefly explaining why it was done the way it was done? Just thinking in terms of someone looking at this in two years and wondering why it's a stream.

andreascful (approved):
Minor nitpicks. Looks great though. Enables processing of large imports! :)

.gitignore
marcolink: anyway - deleted it!

PR #747, opened 2023-02-27
Fix asset transformer with existing protocol

mrjman:
### What does this PR do?
Modifies the regex used in the asset transformer to properly check for protocol on asset url.

### Background Info
The protocol regex was always failing to detect protocol because the case insensitive modifier was being included in the regex instead of being outside the regex definition. This caused every url to have an extra `https:` prepended even those that already provided the protocol.

The test was also incorrectly checking for an extra `https:` to be added and thus started to fail with the changes in this PR so it was updated to pass.

This PR was tested against an import of 300+ images from an external source that all had the protocol already attached. Without the PR the import failed and the validation error showed the extra `https:` being added. With this PR the import worked successfully.


ruderngespra:
Hey, hey! Thanks a lot for this contribution and making us aware of the bug!

PR #830, opened 2023-08-02
chore: update and clean eslint setup

ruderngespra:
When trying to setup `contentful-import` we run into an eslint incompatibility:

image

This PR tries to resolve this dependency by upgrading several eslint related packages.

### changes
- use updated `babel/eslint-parser` plugin instead of deprecated `babel-eslint`
- transpile to only node `18` which is current LTS (instead of 14) (this should also fix PRs like https://github.com/contentful/contentful-import/pull/813)
- update eslint from 7 to 8.
- adjust a test to align with updated eslint


babel.config.json
```diff
@@ -4,13 +4,10 @@
       "@babel/preset-env",
       {
         "targets": {
-          "node": "14"
```
damienxy, line 7: ❤️

damienxy (approved):

ruderngespra:
wrong semantic release commit :( I'll follow up with another PR.

PR #880, opened 2023-09-22
feat: use tsc instead of babel to create bundle [PHX-2872]

ruderngespra:
Starting point for a gradual migration to typescript.

Main point of this PR is to use `tsc` instead of `babel` for bundle creations. As a next step, we might move to something like [tsup](https://tsup.egoist.dev/), but let's go step by step.

### things done in this PR
- Remove ~all~ many things related to babel
- Instead create the bundle with `tsc`
- Add typescript and latest @types/node as dependencies
- Move to `ts-jest` and move related config out of `package.json`
- For `eslint` use typescript-eslint parser instead of babel (new upcoming issues are treated as warnings for now)


marcolink (approved):
I think this is a great initial setup for the TS migration 🎉

PR #886, opened 2023-09-26
feat: add tsup and workaround to export both esm and cjs

ruderngespra:
[This](https://github.com/contentful/contentful-import/pull/880) broke support for common js, this PR fixes it (see the long comment)

- adds tsup as bundler
- run tests synchronously as some of their processes were interfering with each other


marcolink (approved):
🎉

PR #890, opened 2023-09-27
typescript migration part1

ruderngespra:
- Moves all utils files to ts and adjusts a few ts errors
- adjusts build to include `tsc` and tsup won't do actual type checking


lib/utils/sort-locales.ts
```diff
@@ -20,7 +20,7 @@ export default function sortLocales (locales) {
   return sortByFallbackKey(localeByFallback)
 }
 
-function sortByFallbackKey (localeByFallback, key) {
+function sortByFallbackKey (localeByFallback, key?: any) {
```
damienxy, line 23:
```suggestion
function sortByFallbackKey (localeByFallback, key?: string) {
```

lib/utils/validations.ts
```diff
@@ -27,7 +27,7 @@ const assertPayload = (payload) => {
     attachEntityName(result.error.details, payload)
     const invalidEntityCount = countInvalidEntities(result.error.details).join(', ')
     result.error.message = `${invalidEntityCount} - Get further details in the error log file`
-    delete result.error._object
+    delete result.error._original
```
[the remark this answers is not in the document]
ruderngespra: joi adjusted their error object, we are deleting the same thing only that it's now called `_original`, (before it was `_value` and before that it was `_object`) (https://github.com/hapijs/joi/commit/6182a9b13a74286bd5cfcef1481e1c3555569266)

types showed that to me.

lib/utils/sort-locales.ts
marcolink: Should `localeByFallback` also be typed?

marcolink (approved):

PR #891, opened 2023-09-27
chore: adjust syntax and error handling

ruderngespra:
Moves some files in `tasks` and `transform` to typescript and adjust accordingly.

### all changes
- creates an error with filePath property --> `ContentfulAssetError` as the standard Error object doesn't allow for a `filePath` property.
- Adjust types in get-destination-data to make current return value types of `batchedIdQuery` explicit (otherwise ts assumes the resources on `result` are supposed to be arrays and not promises and throws an error.


lib/tasks/get-destination-data.ts
```diff
@@ -57,6 +57,16 @@ function getIdBatches (ids) {
   return batches
 }
 
+type AllDestinationData = {
+  contentTypes: Promise<any[]>
```
marcolink, line 61: These are all the `Array<{EntityName}Props>` from `contentful-management`

PR #892, opened 2023-09-27
chore: add types to push-to-space main data objects [PHX-2879]

ruderngespra:
Again only moving a few more types to document existing behavior, mainly around the data structure used in `push-to-space`.


lib/types.ts
```diff
@@ -0,0 +1,11 @@
+import type { AssetProps, ContentTypeProps, EditorInterfaceProps, EntryProps, LocaleProps, TagProps, WebhookProps } from 'contentful-management'
+
+export type SpaceResources = {
```
marcolink, line 3: To me they look more like environment resources 🤔

ruderngespra: That was also my first thought, [but webhooks aren't environment aware](https://www.contentful.com/faq/environments/#what-entities-can-be-copied-to-a-space-environment). Given the time this tool was created (before the introduction of environments) my suggestion is to stick with the jargon of that time  and only if we actually restructure the logic also rename.

ruderngespra: done ✅

marcolink (approved):

PR #896, opened 2023-09-29
chore(ts): move test files to ts [PHX-2885]

ruderngespra:
This is the first PR that starts with gradually also moving test files to ts.

During the migration I mainly focused on making types conform with the current test expectations, e.g. mockData passed to `pushToSpace`. All the types are of course still very loose with a lot of `any` as I have not yet thought through the whole way of how data is transformed before being pushed to a space.

### main things done in this PR
- Noticed that we have `SourceData` and `TransformedSourceData` and only the latter is being passed to `pushToSpace`. Started adjusted the types accordingly (better naming etc., but still wip).
- In order to conform with the tests, I loosened the type restrictions around `pushToSpace` params, some seem optional, so for now I made them so.
- Had to adjust a lot of types in `pushToSpace.test.ts`, partly just typecasting for now.


lib/types.ts
```diff
@@ -9,3 +9,33 @@ export type Resources = {
...
+  fields: { file: { upload: string, uploadFrom: Link<'Upload'> }[] },
+  sys: {id: string}
+}
+
+export type AssetWithTransformed = {
+  transformed: TransformedAsset,
+  original: any
+}
+
+type OriginalContentType = ContentTypeProps
```
marcolink, line 25: Do we need this type?

marcolink (approved):

PR #908, opened 2023-10-05
feat(ts): add public ts types

ruderngespra:
This adds publicly available and enforced ts interfaces to `runContentfulImport`.

#### TODOS

- Fix ts errors in tests (params partly don't match my docs derived ts expectations) ✅ 
- Expand readme with missing params (see comment) ✅

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -55,7 +55,7 @@ type RunContentfulImportParams = {
   // TODO These properties are not documented in the Readme
   timeout?: number,
   retryLimit?: number,
-  config?: object,
+  config?: string,
 }
 
 async function runContentfulImport (params: RunContentfulImportParams) {
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -32,7 +32,33 @@ function createListrOptions (options) {
   }
 }
 
-async function runContentfulImport (params) {
+// These type definitions follow what is specified in the Readme
+type RunContentfulImportParams = {
+  spaceId: string,
+  environmentId?: string,
+  managementToken: string,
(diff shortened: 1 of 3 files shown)


mayakarabula (approved):

test/unit/index.test.ts
```diff
@@ -94,7 +94,9 @@ test('Stops import when default locales does not match', async () => {
             default: true
           }
         ]
-      }
+      },
+      spaceId: 'SPACE_ID',
```
marcolink, line 98: Should we use the same format for both params?

marcolink (approved):
Great 🎉 
Since we are touching the readme already - should we fix the broken badges?

lib/index.ts
```diff
@@ -32,7 +32,33 @@ function createListrOptions (options) {
...
+  skipContentModel?: boolean,
+  skipLocales?: boolean,
+  skipContentPublishing?: boolean,
+  uploadAssets?: boolean,
+  assetsDirectory?: string,
+  host?: string,
+  proxy?: string,
+  rawProxy?: string,
+  rateLimit?: number,
+  headers?: object,
```
Vinz93, line 52: Maybe we can be more specific here
```suggestion
  headers?: [header: string]: string | string[] | undefined,
```

mayakarabula: that's a good idea! the shape is a bit different tho, I will update it
https://github.com/contentful/contentful-import/blob/feat/adjust-types-for-public-interface/lib/utils/headers.ts#L13

Vinz93 (approved):

Issue #599, opened 2022-02-04
Importing Editor Interfaces ignores sidebar and editors


aghuddleston:
Currently the import code for the editor interfaces copies over

```
            ctEditorInterface.controls = editorInterface.controls
            ctEditorInterface.groupControls = editorInterface.groupControls
            ctEditorInterface.editorLayout = editorInterface.editorLayout
```

But, as Contentful has added more options (thank you) for configuring, there are more editor interface configurations (https://contentful.github.io/contentful-management.js/contentful-management/7.53.0/interfaces/EditorInterface.html)  such as

```
sidebar
editors
```

It would be nice if the import included sidebar and editors in the Editor Interfaces import.

PR #915, opened 2023-10-10
feat: import sidebar configuration [ZEND-4192]

marcolink:
Respect more EditorInterface props when importing.  Adds `sidebar` and `editors` to import routine. 

Solves #599


ruderngespra (approved):
🚢

Issue #62, opened 2017-09-21
Bad Request: "You need to omit a field before deleting it"


wiledal:
Hey,

I cannot import a space that has deleted fields in a content model.

```
info import Contentful Import Tool
info import Importing data ...
info Checking if destination space already has any content and retrieving it
info Pushing content to destination space
ERR! import BadRequest: {
ERR! import   "status": 400,
ERR! import   "statusText": "Bad Request",
ERR! import   "message": "You need to omit a field before deleting it",
```

Basically, we need to "Disable in response" before we can delete a field.

I get that this might be a design choice when it happens in the visual CMS, but it makes properly syncing spaces impossible.

Could we add a `--force` flag to allow "insecure updates" to happen?

axe312ger:
There is no option to enable insecure updates in our API.

To properly migrate your content types, we just released https://github.com/contentful/migration-cli which takes care of omitting fields before deleting them.

I recommend you to use the `--skip-content-model` to import content and use the migration-cli to do your field transformations.

axe312ger:
Okay, we discussed this and decided to integrate omitting of the fields to this tool as well. I still recommend the migration-cli for more complex usage, but expect a feature release for the import tools soon :)

wiledal:
Hey, that migration tool looks powerful, but building the models becomes a less intuitive process. 
We enjoy using the visual CMS to build the content model for rapid results.

Thanks!

wiledal:
Any update on this? Would be very useful to us currently.

floelhoeffel:
Hey @wiledal - we know with the current offering working with multiple environments is a bit tricky. However we are about to release some new features on the API side which will make this a lot easier soon. Another thing we are considering for the migrations tooling is to support a workflow where you can generate a migration script from an existing content model.

In the meantime, we recommend to manually remove the field in the target space from the CT and try the import again. Import and export are not designed to write into non-empty spaces, it works in many cases but there is no logic in place to resolve conflicts. 

Hope this helps

floelhoeffel:
The reason why we don't want "merging/syncing" functionality in the import export tooling is that there is so many side effects. The only way to handle that fail save is through a migration script. As mentioned in my previous message, generating a migration script from a web app curated content model is a very interesting use case we are considering to add to the tooling.

dnlmzw:
Any update on this? Also curious as the dev, staging, production cycle gets a little cumbersome at the moment.

lukearmstrong:
[...]

PR #919, opened 2023-10-10
feat: add force flag to import omitted fields [#62]

marcolink:
Introduce a new flag `force` to enforce content type imports with omitted fields. If the flag is present, we delete omitted fields before importing the content type. 

Addresses the feature request formulated in several issues:
- solves https://github.com/contentful/contentful-cli/issues/74
- https://github.com/contentful/contentful-import/issues/62


lib/index.ts
```diff
@@ -128,8 +132,12 @@ async function runContentfulImport (params: RunContentfulImportParams) {
     {
       title: 'Apply transformations to source data',
       task: wrapTask(async (ctx) => {
-        const transformedSourceData = transformSpace(ctx.sourceDataUntransformed, ctx.destinationData)
-        ctx.sourceData = transformedSourceData
+        const customTransformers: Partial<typeof transformers> = {}
+        if (options.force) {
+          customTransformers.contentTypes = (contentType: ContentTypeProps) => pipe(transformers.contentTypes, forceDeleteOmittedFieldTransform)(contentType)
```
ruderngespra, line 137: Personally not the biggest fan of adding the pipe dependency "just" for this, not sure about the benefit, we also just can pass the params through all all transformers instead, no?

lib/transform/transformers.ts
```diff
@@ -54,21 +60,34 @@ export function assets (asset, _, tagsEnabled = false) {
...
     // This will implicitly remove the locale ID
     // which then causes the create path to not pick `createLocaleWithId` but `createLocale` instead
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
     transformedLocale.sys = pick(destinationLocale.sys, 'id')
   }
 
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
   return transformedLocale
```
ruderngespra, line 76: why do we need this ignore here, what's the issue?

test/unit/transform/force-delete-omitted-field-transform.test.ts
```diff
@@ -0,0 +1,11 @@
+import { cloneMock } from 'contentful-batch-libs/test/mocks/'
+import { ContentTypeProps } from 'contentful-management'
+import { forceDeleteOmittedFieldTransform } from '../../../lib/transform/force-delete-omitted-field-transform'
+
+test('It should transform webhook by dropping omitted fields', () => {
```
ruderngespra, line 5: why webhook?

ruderngespra (approved):
Very cool stuff, love all the additional types as well. Maybe we can add an additional entry to the readme list [here](https://github.com/contentful/contentful-import/blob/master/README.md#gear-configuration-options).

Also, what do you think about "delete-omitted-fields" instead of "force", just because force could be needed for other things in the future and sounds quite generic.

lib/transform/transformers.ts
marcolink: The retuned shape only has a subset of data in the `sys` field (only the `id` field)

test/unit/transform/force-delete-omitted-field-transform.test.ts
marcolink: @ruderngespra good catch - changed it 🙏

PR #950, opened 2023-11-08
fix: adopt new batch-lib version [NONE]

marcolink:
Align with the changes from the new major version contentful-batch-libs@10

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -7,7 +7,7 @@ import VerboseRenderer from 'listr-verbose-renderer'
 import { startCase } from 'lodash'
 import PQueue from 'p-queue'
 
-import { displayErrorLog, setupLogging, writeErrorLogFile } from 'contentful-batch-libs/dist/logging'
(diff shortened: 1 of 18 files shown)


lib/index.ts
```diff
@@ -193,8 +191,8 @@ async function runContentfulImport (params: RunContentfulImportParams) {
       })
     })
     .then((data) => {
-      const errorLog = log.filter((logMessage) => logMessage.level !== 'info' && logMessage.level !== 'warning')
-      const displayLog = log.filter((logMessage) => logMessage.level !== 'info')
+      const errorLog = log.filter(isErrorLog)
+      const displayLog = log.filter(isDisplayLog)
```
Brad-Turner, line 195: `log` only accepts messages with levels `error` or `warn`, I think this filter does nothing?

Brad-Turner (commented):
Looks good to me. Would be nice to reduce the reliance on lodash 🤔

PR #1071, opened 2024-03-06
fix: remove ANSI color codes from table output for better CI/CD compat [ZEND-4747]

jjolton-contentful:
see https://github.com/cli-table/cli-table3/blob/master/basic-usage.md#basic-usage---disable-colors---used-often-in-the-examples-and-tests and https://github.com/contentful/contentful-export/pull/1711


t-col (approved):
✨

PR #1344, opened 2025-02-20
fix(bundling): split esm and cjs

itxch:
## Changes
- Set `splitting: true` in the `tsup` config, so that the library is built with separate `esm` and `cjs`

This fixes issues with importing the library where I get an error "can't find variable: module"
CleanShot 2025-02-20 at 22 34 48

### Other Changes
- Set nvmrc to node v20, as thats what actually required to build the project
- Change the order of exports to remove `tsup` warnings


itxch:
This seems to introduce a bunch more issues, due to lodash not properly supporting ESM. Will close for now and re-open

PR #1345, opened 2025-02-21
feat(bundling): proper esm support

itxch:
Adds proper ESM support, mainly by ensuring all dependancies play nicely with ESM. This also maintains backwards CJS compatibility. But as it updates to LTS version of node (22), **I'd recommend this gets released as a breaking change.**

Depends on https://github.com/contentful/contentful-management.js/pull/2556. Once this PR is merged we'll need to bump the version of `contentful-management.js` used.

## Main Changes
1. Update version of node to v22. Also add [volta](https://volta.sh/) entry to package.json
 to allow for easy version management
2. Set package to use `ESM` by default.
3. Update dependancies to ESM compatible versions.
4. Replace `tsup` with `vite` for modern bundling and transpiling features.
5. Replace `jest` with `vitest` due to native ESM / TypeScript support.
6. Update `README.md` and `CONTRIBUTING.md` to reflect the current state of the repo
7. Update CircleCI config to use the correct version of node.

## Other

- package.json scripts have been updated.
- Added `example.test.env` file to show what values are needed to locally run tests.

The reason there is alot of change is due to:
1. Updating a bunch of the imports across the repo
2. Updating tests to use vitest rather than jest.


t-col:
hi @itxch - thank you for the contribution, but unfortunately I will closing this PR, as it is too large, and contains too many changes, in a single changeset for our standards. Generally we would consider a PR only addressing a single change at a time, however, we are not interested in supporting any of the above changes at this time. That is, we will not be changing the supported node version until v18 is deprecated, we will not be making the package ESM by default at this time (but it will be optional), and we will not be moving to vite/vitest at this time either.

Best,
Tyler

itxch:
Created an issue: https://github.com/contentful/contentful-import/issues/1350 to continue the conversation

PR #1387, opened 2025-04-17
fix(upload-assets param): remove the "default: false"

ethan-ozelius-contentful:
# Error 1

### Description
Providing a default param (of any type/kind) results in that param existing (aka being set), which triggers the 
implies("upload-assets", "assets-directory") check, which obviously fails when neither param is passed by user.

```shell
npx contentful-import \
  --space-id "$CONTENTFUL_SPACE_ID" \
  --management-token  "$CONTENTFUL_PRODUCTION_ACCESS_TOKEN" \
  --content-file contentful-export-<contentful-space-id>-master-2025-04-16T16-40-58.json
```

results in the following error

```
Missing dependent arguments:
 upload-assets -> assets-directory
```

### Root cause
`--upload-assets` `default: false` option unintentionally causes the `implies("upload-assets", "assets-directory")` statement to kick in.

### Solution
remove the `default: false`.

# Error 2
### Description
Running `contentful-import` as a CLI executable does NOT properly spread the passed params.

```shell
npx contentful-import \
  --space-id "$CONTENTFUL_SPACE_ID" \
  --management-token  "$CONTENTFUL_PRODUCTION_ACCESS_TOKEN" \
  --content-file contentful-export-<contentful-space-id>-master-2025-04-16T16-40-58.json
```

results in the following error

```
Error: The `spaceId` option is required.
```

### Root cause
`usageParams.ts` is responsible for gathering all the options to be passed to `runContentfulImport()`, which is gathers from potentially `~/.contentfulrc.json` and from the args passed in the CLI.  The problem is that [this line](https://github.com/contentful/contentful-import/blob/main/lib/parseOptions.ts#L42) is spreading on an object like this: `{ default: { /* all the cli params */ } }`

resulting in a payload like this:

```json
{
    "skipContentModel": false,
    "skipLocales": false,
    "skipContentPublishing": false,
    "skipAssetUpdates": false,
    "skipContentUpdates": false,
    "useVerboseRenderer": false,
    "environmentId": "master",
    "rawProxy": false,
    "uploadAssets": false,
    "rateLimit": 7,
    "default": { // this should not be wrapped in a default!
        "_": [],
        "space-id": "abcedf2847",
        "spaceId": "abcedf2847",
        "management-token": "bsdf",
        "managementToken": "bsdf",
        "content-file": "asdf",
        "contentFile": "asdf",
        "environment-id": "master",
        "environmentId": "master",
        "content-model-only": false,
        "contentModelOnly": false,
        "skip-content-model": false,
        "skipContentModel": false,
        "skip-locales": false,
        "skipLocales": false,
        "skip-content-publishing": false,
        "skipContentPublishing": false,
        "host": "api.contentful.com",
        "raw-proxy": false,
        "rawProxy": false,
        "rate-limit": 7,
        "rateLimit": 7,
        "$0": "bin/contentful-import"
    },
    "headers": {
        "CF-Sequence": "11ca8fd7-322d-412f-b89c-e39eed5ac7d2"
    }
}
```

### Solution
To maintain backwards compatability with using this tool both as a CLI executable, as well as a regular old js module, update the bin statement to import the default

```
- const usageParams = require('../dist/usageParams')
+ const usageParams = require('../dist/usageParams').default
```

dx-48


lib/usageParams.ts
ethan-ozelius-contentful, line 1: FYI, I would've loved to add unit tests, but there isn't an existing unit test to test the `yargs` inputs specifically, just the consumer of these yargs.  One option could be to install/implement [nixt](https://www.npmjs.com/package/nixt) which is also used in `contentful-cli`, but that felt like overkill just to make sure that yargs is working correctly.

open to suggestions

lib/usageParams.ts
```diff
@@ -48,7 +48,6 @@ export default yargs
   .option('upload-assets', {
     describe: 'Uses local asset files and uploads them instead of pointing to the URLs of previously uploaded assets. Requires assets-directory',
     type: 'boolean',
-    default: false
```
elylucas, line 51: Im wondering why the check for all the required args would go into effect when default is false here. It feels a bit wrong to break the pattern that is set for the other params by removing `default: false` and instead look into why it triggers when it is defined.

ethan-ozelius-contentful: If I'm understanding [this yargs PR](https://github.com/yargs/yargs/pull/1985) correctly, this is intentional.  The check is for the _existence of_, not the value of.


```js
// check if key/value exists
- val = argv[val];
+ val = Object.prototype.hasOwnProperty.call(argv, val);
```

whitelisab (dismissed):
LGTM ⭐

README.md
```diff
@@ -75,6 +75,23 @@ contentfulImport(options)
...
 
+## 💾 Usage as a CLI executable
+
+1. Build the package:
+
+```bash
+$ npm run build
+```
+
+2. Execute the CLI tool by running the binary and passing the required parameters:4
```
whitelisab, line 86: Should the `4` be removed?

```suggestion
2. Execute the CLI tool by running the binary and passing the required parameters:
```

whitelisab (dismissed):
one typo question on the `README`, otherwise approving!

whitelisab (approved):

PR #1462, opened 2025-08-20
fix: use filename for local asset to match export behavior [DX-336]

whitelisab:
### PR Description
When the contentful-import tool is used with the `uploadAssets` option, the tool finds the assets in the local file system and uploads them. 

We are proposing changes to where these files are saved when using the `contentful-export` tool (see https://github.com/contentful/contentful-export/pull/2119), and this PR adds matching logic for the filename the tool should look for in the local file system. The tool will look for the original filename in the local file system, rather than one extracted from the `url` property. In the case where the filename is not available, the fallback is to use the decoded filename from the url path instead of the encoded one. Tests have been added to reflect this new logic.

See related issue: https://github.com/contentful/contentful-export/issues/2113 and [PR](https://github.com/contentful/contentful-export/pull/2119).

### DIFF
--- a/test/unit/tasks/push/assets.test.ts
+++ b/test/unit/tasks/push/assets.test.ts
@@ -19,7 +19,10 @@ jest.mock('fs')
 
 const assetPaths = [
   'assets/images/contentful-en.jpg',
-  'assets/images/contentful-de.jpg'
+  'assets/images/contentful-de.jpg',
+  'assets/images/测试文件.jpg',
+  `assets/images.ctfassets.net/${'测试文件'.repeat(10)}.jpg`,
+  'assets/images/different filename.jpg'
 ]
 
 let requestQueue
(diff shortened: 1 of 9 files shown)


test/unit/tasks/push/assets.test.ts
```diff
@@ -154,16 +159,59 @@ test('Process assets fails', async () => {
...
   const createReadStreamSpy = jest.spyOn(fs, 'createReadStream')
   const fileUrl = 'https://images/contentful-en.jpg'
-  const stream = await getAssetStreamForURL(fileUrl, 'assets')
+  const stream = await getAssetStreamForURL(fileUrl, 'assets', 'contentful-en.jpg')
 
   expect(createReadStreamSpy).toHaveBeenCalledTimes(1)
   expect(stream instanceof Stream).toBe(true)
 })
+
+test('Get asset stream for url: Handle Unicode filename', async () => {
```
elylucasctfl, line 176: nice additions to the tests

elylucasctfl (approved):
Code looks good. As long as theres no concerns with backwards compat i brought up in Slack, then gtg!

PR #1515, opened 2025-11-11
chore: update CI workflows to GHA and publish to NPM

michaelphamcf:
- Migrate to GHA for CI workflows
- Use updated Contentful practice for OSS workflows
- Semantic-release publish packages to NPM using OIDC
- Lint job commented out until ready to be part of the pipeline

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "bluebird": "^3.7.2",
         "cli-table3": "^0.6.5",
         "contentful-batch-libs": "^9.6.0",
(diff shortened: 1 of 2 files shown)


.github/workflows/check.yaml
```diff
@@ -0,0 +1,63 @@
+name: Run Checks
+
+on:
+  workflow_call:
+
+jobs:
+  # lint:
```
elylucasctfl: did we need to uncomment the lint job to get it back in? If not, put in a note to say why its commented out.

PR #1635, opened 2026-05-24
fix: correct processingCheckRetry typo to processingCheckRetries

### LINKED ISSUE #1632: [BUG] retryLimit is silently ignored due to parameter name typo, causing indefinite hangs on asset processing
## Bug Description

Two related bugs cause the `retryLimit` option to be completely non-functional and asset
imports to hang indefinitely with no recovery when Contentful's process endpoint does not
respond.

**Bug 1:** `retryLimit` is silently ignored because `contentful-import` passes
`processingCheckRetry` (singular) to `processForLocale`, but `contentful-management`
expects `processingCheckRetries` (plural). The unknown key is discarded and the SDK falls
back to its default of 10 retries regardless of the configured value.

**Bug 2:** Even if Bug 1 were fixed, there is no HTTP-level timeout on the PUT request
that triggers asset processing. If Contentful's API does not respond to that request, the
promise never resolves, the polling loop never starts, and the import hangs indefinitely —
`retryLimit` and `timeout` cannot help because they only govern the polling loop, not the
triggering request.

## Steps to Reproduce

**Bug 1 (parameter name typo):**

1. Install `contentful-import` and call it programmatically with `retryLimit: 1`
2. Observe that the SDK uses 10 retries (its internal default) rather than 1
3. Confirm by inspecting `dist/index.js` lines 299–300:

```js
retryLimit && { processingCheckRetry: retryLimit }  // singular — unknown to the SDK
```
vs. contentful-management dist/cjs/index.cjs line 1089:
```js
options: { processingCheckRetries, processingCheckWait }  // plural — what the SDK expects
```
Bug 2 (indefinite hang):

1. Run contentful space import (CLI or programmatic API) against a space containing PDF
assets or certain large images
2. Wait for the "Uploading Assets" phase to complete
3. Observe the "Importing Assets" phase — for certain assets the spinner stops with no
output, no error, no rate-limit warning
4. The process does not recover regardless of --retry-limit or --timeout values

## Expected Behavior

- Setting retryLimit: 1 should cause the importer to give up on a stuck asset after 1
polling attempt (~3 seconds) and move on to the next asset, logging the failure
- When the process endpoint does not respond, the import should time out after the
configured timeout / retryLimit window, log the failed asset, and continue rather
than hanging indefinitely

## Actual Behavior

- retryLimit has no effect — the SDK always uses its default of 10 retries × 3000ms =
30 seconds because the key name passed is wrong
- When Contentful's process endpoint does not return a response, the import hangs
indefinitely. No timeout fires. The only recovery is to kill the process manually.
Observed consistently for PDFs of all sizes and some image assets

## Code Sample

Bug 1 — contentful-import dist/index.js lines 293–300

```js
// processingCheckRetry (singular) is passed, but contentful-management expects
// processingCheckRetries (plural). The key is silently ignored.

async function processAssets({ assets, timeout, retryLimit }) {
const processingOptions = Object.assign(
    {},
    timeout && { processingCheckWait: timeout },
    retryLimit && { processingCheckRetry: retryLimit } // ← BUG: should be processingCheckRetries
)
// processingOptions is passed to asset.processForLocale(locale, processingOptions)
// The SDK destructures: { processingCheckRetries, processingCheckWait }
// processingCheckRetry is unknown and discarded — SDK uses its default of 10
}

// Bug 2 — processForLocale makes an HTTP PUT to trigger processing.
// If the API does not respond, the promise never resolves.
// processingCheckRetries only governs the polling loop that runs AFTER the PUT returns —
// it never fires if the PUT itself hangs.
await runContentfulImport({
spaceId: 'your-space-id',
managementToken: 'your-token',
contentFile: 'export.json',
uploadAssets: true,
assetsDirectory: './backup',
retryLimit: 1,  // has no effect due to Bug 1
timeout: 5000,  // also has no effect on the hanging PUT request
})
// Import hangs indefinitely on certain assets with no output
```

## Environment

- OS: Linux (Ubuntu 22.04)
- Package Version: contentful-import 10.0.5 (also confirmed in 10.0.6 source on main)
- contentful-management: 12.3.2
- Node Version: 22.x
- Package Manager: pnpm 10.x

## Error Messages/Logs

✔ Uploading Assets (2686s)
⠋ Importing Assets
→ ✔ Processing Asset some-asset-name (assetId)
Publishing Assets
...
[No further output — process hangs indefinitely]

When the process endpoint eventually does respond (rare), the error seen is:

AssetProcessingTimeout: Asset is taking longer than expected to process.

In most cases however no error is thrown — the HTTP request simply never returns.

## Screenshots

N/A

## Additional Context

The hang affects PDFs of all sizes (from 200 KB form documents to 20 MB magazine PDFs)
and some image assets. The affected assets are valid files — they import successfully
through the Contentful web UI. The issue is specific to the management API processing
path when triggered via contentful-import.

Related issues:
- contentful-cli #2073 (https://github.com/contentful/contentful-cli/issues/2073) — PDF asset hangs indefinitely during import
- contentful-management.js #1516 (https://github.com/contentful/contentful-management.js/issues/1516) — AssetProcessingTimeout with no resolution

We worked around Bug 2 by bypassing contentful-import for asset processing and using
fetch() with AbortController directly against the process endpoint:

```
const controller = new AbortController()
setTimeout(() => controller.abort(), 30_000)

await fetch(
`https://api.contentful.com/spaces/${spaceId}/environments/${envId}/assets/${assetId}/files/${locale}/process`,
{
    method: 'PUT',
    headers: {
    Authorization: `Bearer ${token}`,
    'X-Contentful-Version': String(version),
    },
    signal: controller.signal,
}
)
```

This reliably times out and moves on from stuck assets rather than hanging indefinitely.

## Possible Solution

Bug 1 is a one-character fix:

- retryLimit && { processingCheckRetry: retryLimit }
+ retryLimit && { processingCheckRetries: retryLimit }

Bug 2 requires either:

- Passing an HTTP request timeout when creating the internal contentful-management
client inside contentful-import (e.g. via an axios timeout option), or
- Wrapping the processForLocale call in a Promise.race against a configurable
deadline, so that assets whose process endpoint call never returns are logged as
failures and skipped rather than blocking the entire import

franko-c:
Fixes #1632

## Summary

`lib/tasks/push-to-space/assets.ts:70` passes `retryLimit` under the key `processingCheckRetry`, but the Contentful Management SDK expects `processingCheckRetries` (plural). one-character typo that causes the retry limit to be silently ignored during asset processing.

## Changes

- `lib/tasks/push-to-space/assets.ts`: `processingCheckRetry` → `processingCheckRetries`

## Verification

targeted harness confirmed the fix — baseline (singular form) fails, patched (plural form) passes.

### DIFF
--- a/lib/tasks/push-to-space/assets.ts
+++ b/lib/tasks/push-to-space/assets.ts
@@ -67,7 +67,7 @@ export async function processAssets ({
   const processingOptions = Object.assign(
     {},
     timeout && { processingCheckWait: timeout },
-    retryLimit && { processingCheckRetry: retryLimit }
+    retryLimit && { processingCheckRetries: retryLimit }
   )
 
   const pendingProcessingAssets = assets.map(async (asset) => {


franko-c:
thanks @ethan-ozelius-contentful — closing since the typo fix shipped in #1638 🎉

PR #1665, opened 2026-07-09
chore: improve logging and error handling [AIS-168]

ethan-ozelius-contentful:
## Summary
1. raise/log a message when asset processing fails - see `lib/tasks/push-to-space/assets.ts`
  - was previously failing silently 🫢 
3. specifically handle VersionMismatch errors - see `lib/tasks/push-to-space/creation.ts`
  - was previous being treated like a generic error.

### DIFF
--- a/lib/tasks/push-to-space/creation.ts
+++ b/lib/tasks/push-to-space/creation.ts
@@ -192,8 +192,9 @@ function createTagInDestination (context, sourceEntity) {
 
 /**
  * Handles entity creation errors.
- * If the error is a VersionMismatch the error is thrown and a message is returned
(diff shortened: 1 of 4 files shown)


lib/tasks/push-to-space/creation.ts
```diff
@@ -203,6 +204,12 @@ function handleCreationErrors (entity, err) {
       return entity
     }
   }
+
```
ethan-ozelius-contentful, line 207: Specifically handle a situation where the entity to import already exists and is newer than the one being imported.

tylerpina (approved):

PR #1668, opened 2026-07-13
fix: exo - Added destination data prefetching for ExO entities, use version-aware create-or-update instead of blind upserts [AIS-130]

ethan-ozelius-contentful:
`get-destination-data.ts:`
- Added ComponentTypeProps, DataAssemblyProps, ExperienceProps, FragmentProps, TemplateProps to the import
- Added fetchAllExoEntities — a cursor-paginated fetch helper for the ExO APIs (which use pages.next rather than offset/limit)
- Extended AllDestinationData with optional fields for all 5 ExO entity types, plus a TODO comment for DesignTokens
- Added plainClient, includeExperienceOrchestration to GetDestinationDataParams
- When includeExperienceOrchestration is true, fetches all destination componentTypes, templates, fragments, dataAssemblies, and experiences via plainClient, with a TODO for designTokens

`index.ts:`
- Threads plainClient and includeExperienceOrchestration through to getDestinationData

`push-to-space.ts:`
- All 5 ExO entity tasks now look up destinationDataById before acting: if the entity exists in the destination, passes the destination sys.version in the payload (satisfying the API's
optimistic concurrency requirement); if not, passes the payload without sys
- Removed the debug console.log statements from the ComponentTypes task
- Removed the inner try/get-then-create for DataAssemblies (now handled uniformly via the pre-fetched map)
- Added a TODO comment for the DesignTokens task

### Unit tests

  test/unit/tasks/push/push-to-space-exo.test.ts — 11 tests across 5 entity types (CT, Template, Fragment, DataAssembly, Experience):
  - CREATE path: verifies upsert/create is called with no sys in the payload when the entity is absent from the destination map
  - UPDATE path: verifies the destination entity's sys.version is threaded into the payload (the optimistic concurrency value), and that DataAssembly specifically uses update not create
  - Skip guard: verifies no calls are made when includeExperienceOrchestration: false

  test/unit/tasks/get-destination-data-exo.test.ts — 9 tests:
  - Verifies cursor pagination (pageNext) is used for ExO entities, offset pagination is used for standard entities
  - Verifies ExO methods are not called when the flag is off
  - Verifies multi-page cursor traversal (150 items across 2 pages → 2 getMany calls)
  - One test per entity type confirming the returned data is populated from the mock
  - Empty-destination case (all arrays come back length 0)

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -121,11 +121,13 @@ async function runContentfulImport (params: RunContentfulImportParams) {
       task: wrapTask(async (ctx) => {
         const destinationData = await getDestinationData({
           client: ctx.client,
+          plainClient: ctx.plainClient,
           spaceId: options.spaceId,
           environmentId: options.environmentId,
           sourceData: options.content,
           skipLocales: options.skipLocales,
           skipContentModel: options.skipContentModel,
+          includeExperienceOrchestration: options.includeExperienceOrchestration,
           requestQueue
         })
 
--- a/lib/tasks/get-destination-data.ts
+++ b/lib/tasks/get-destination-data.ts
@@ -1,7 +1,7 @@
 import Promise from 'bluebird'
 
 import { logEmitter } from 'contentful-batch-libs/dist/logging'
-import type { AssetProps, ContentTypeProps, EntryProps, LocaleProps, TagProps, WebhookProps } from 'contentful-management'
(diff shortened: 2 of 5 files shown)


michaelphamcf (approved):
Maybe worth it to add some tests for the new logic?

PR #1669, opened 2026-07-13
feat: implement URN rewriting transformer for ExO entities [AIS-128]

ethan-ozelius-contentful:
## Context
```
⏺ Why URN Rewriting Is Only Needed for ExO Entities

  The root cause: how ExO entities reference each other vs. how standard entities do.

  Standard Contentful entities (entries, assets, content types, etc.)

  These reference each other via sys.type: 'Link' objects that look like:

  { "sys": { "type": "Link", "linkType": "Entry", "id": "abc123" } }

  The id is just an opaque string with no space/environment baked in. At query time, the API resolves that link within the context of the current space/environment. So when you import into a
  new space and preserve IDs (which contentful-import does), these links just work — they point to whatever entity with that ID exists in the destination space.

  ExO entities (ComponentTypes, Fragments, Templates, DataAssemblies, Experiences, DesignTokens)

  These use a newer reference type: ResourceLink. It embeds the full spatial coordinates directly in the URN:

  crn:contentful:::experience:spaces/{spaceId}/environments/{envId}/{entityType}/{entityId}

  The space and environment IDs are hardcoded inside the reference itself. This is what enables ExO's cross-space linking capability (an Experience in space A can reference a ComponentType in
  space B). But it's also what breaks import: if you export from spaces/AAA/environments/BBB and import into spaces/XXX/environments/YYY, every single URN still points back to the source
  space/env. The entity IDs at the end of the URN are preserved on import, but the spaces/.../environments/... segment needs to be rewritten to the destination.

  Why not cross-space-only?

  You might think "well, for a round-trip import to the same space, the URNs are already correct." True — the rewriter would just replace identical values and the result is a no-op. But the
  implementation still needs to handle the general case (cross-space), and it has to be applied to all ExO entities consistently regardless. Also, same-space doesn't mean same-environment —
  you might import master→staging within the same space, which also requires rewriting the environmentId segment.

  TL;DR

  It's not that ExO entities are conceptually different from other Contentful entities in terms of import semantics — it's specifically that they use ResourceLink URNs that embed source
  coordinates, while standard entities use portable ID-only links. The URN rewriting transformer exists solely to handle that encoding difference.

```

## Summary

- Adds `rewriteUrns()` helper in `lib/transform/transformers.ts` that recursively walks an ExO entity and rewrites any `ResourceLink` URN's `spaces/{src}/environments/{src}` segment to the destination space/environment
- Wires up six ExO entity transformers (`componentTypes`, `templates`, `fragments`, `dataAssemblies`, `experiences`; `designTokens` is a passthrough since it has no URN references)
- Extends `transformSpace()` to accept a `TransformContext` (destination space + env IDs) and thread it into ExO transformers; ExO entities are auto-included when present in source data
- Passes destination coordinates from `runContentfulImport` options into `transformSpace`

## Why ExO only

Standard Contentful entities reference each other via `sys.type: 'Link'` with a bare ID — no space/env baked in, so they resolve correctly in any destination. ExO entities use `ResourceLink` URNs that embed the full `spaces/{id}/environments/{id}` path, which is what enables cross-space linking but also what breaks on import without rewriting.

## Acceptance criteria

- ✅ Cross-space import: URNs point to destination space/env, entity IDs unchanged
- ✅ Round-trip export→import on same space: URNs unchanged (no-op)
- ✅ `designTokens` passes through unchanged

## Test plan

- [ ] Unit tests for `rewriteUrns` (bare link, nested, same-space no-op, entity ID preservation) — added in `test/unit/transform/transformers.test.ts`
- [ ] Unit test for `componentTypes` transformer with context
- [ ] Unit test for `designTokens` passthrough
- [ ] `npm run test:unit` passes (99/99)

## Jira

[AIS-128](https://contentful.atlassian.net/browse/AIS-128)

[AIS-128]: https://contentful.atlassian.net/browse/AIS-128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -137,9 +137,13 @@ async function runContentfulImport (params: RunContentfulImportParams) {
     {
       title: 'Apply transformations to source data',
       task: wrapTask(async (ctx) => {
-        const transformedSourceData = transformSpace(ctx.sourceDataUntransformed, ctx.destinationData, undefined, undefined, {
-          destinationSpaceId: options.spaceId,
-          destinationEnvironmentId: options.environmentId ?? 'master'
+        const transformedSourceData = transformSpace({
+          sourceData: ctx.sourceDataUntransformed,
+          destinationData: ctx.destinationData,
+          ctx: {
+            destinationSpaceId: options.spaceId,
+            destinationEnvironmentId: options.environmentId ?? 'master'
+          }
         })
         ctx.sourceData = transformedSourceData
       })
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -137,7 +137,10 @@ async function runContentfulImport (params: RunContentfulImportParams) {
     {
       title: 'Apply transformations to source data',
       task: wrapTask(async (ctx) => {
-        const transformedSourceData = transformSpace(ctx.sourceDataUntransformed, ctx.destinationData)
+        const transformedSourceData = transformSpace(ctx.sourceDataUntransformed, ctx.destinationData, undefined, undefined, {
+          destinationSpaceId: options.spaceId,
+          destinationEnvironmentId: options.environmentId ?? 'master'
+        })
         ctx.sourceData = transformedSourceData
       })
     },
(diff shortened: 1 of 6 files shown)


ethan-ozelius-contentful:
Declining this PR for now, since discussed internally we will not be transforming URNs on import and instead trust the export to manage it's own resource links

PR #1670, opened 2026-07-13
feat: migrate CMA client from legacy to plain adapter [AIS-212]

ethan-ozelius-contentful:
## Context
```
---
  Plan: Migrate to CMA Plain Client

  What changes

  The legacy adapter returns enriched entity objects with instance methods (.publish(), .update(), .archive(), .processForLocale()). The plain adapter returns bare data objects — all
  operations become namespaced calls on the client itself (e.g., client.entry.publish({ spaceId, environmentId, entryId, version })).

  This requires threading client, spaceId, and environmentId through the call stack.

  ---
  File-by-file changes

  lib/tasks/init-client.ts
  - Switch { type: 'legacy' } → { type: 'plain' }, type the return as PlainClientAPI

  lib/tasks/get-destination-data.ts
  - Remove client.getSpace() / space.getEnvironment() — the plain client doesn't need environment objects
  - Replace environment[method](params) dynamic dispatch with explicit calls: client.contentType.getMany(), client.entry.getMany(), client.asset.getMany(), client.locale.getMany(),
  client.tag.getMany()
  - Pass { client, spaceId, environmentId } through the batched query helpers instead of environment

  lib/tasks/push-to-space/push-to-space.ts
  - Remove the "Connecting to space" task (no more getSpace/getEnvironment call needed to get entity objects)
  - Replace ctx.environment.getEditorInterfaceForContentType(id) → client.editorInterface.get({ spaceId, environmentId, contentTypeId })
  - Replace ctEditorInterface.update() → client.editorInterface.update({ spaceId, environmentId, contentTypeId }, data)
  - Replace ctx.environment.createUpload() → client.upload.create({ spaceId, environmentId }, data)
  - Update creation context from { type, target } to { type, client, spaceId, environmentId }

  lib/tasks/push-to-space/creation.ts
  - Change PushToSpaceContext from { type, target } to { type, client, spaceId, environmentId }
  - Replace dynamic target[create${type}WithId]() dispatch with explicit plain client calls per type
  - Replace destinationEntity.update() with type-dispatched client.contentType.update() / client.entry.update() / etc.
  - Remove getPlainData()'s .toPlainObject() branch — plain client already returns plain objects

  lib/tasks/push-to-space/publishing.ts
  - Change signatures to accept { entities, client, spaceId, environmentId, requestQueue }
  - Replace entity.publish() with client[entityNs].publish({ spaceId, environmentId, id, version }) dispatched on entity.sys.type
  - Replace entity.archive() similarly
  - Remove guards checking entity.publish / entity.archive existence

  lib/tasks/push-to-space/assets.ts
  - Change processAssets() to accept client, spaceId, environmentId
  - Replace asset.processForLocale(locale, opts) → client.asset.processForLocale({ spaceId, environmentId, assetId, locale }, opts)

  ---
  Tests

  Unit tests mock these entity objects directly (e.g. { sys: {...}, publish: jest.fn() }). Those mocks will need to be updated to instead mock the plain client namespace methods.
```

## Summary

- Replaces `createClient(config, { type: 'legacy' })` with `{ type: 'plain' }` in `init-client.ts`
- Removes all chained entity object usage — `.publish()`, `.archive()`, `.update()`, `.processForLocale()` instance methods are replaced with namespaced plain client calls (`client.entry.publish()`, `client.asset.archive()`, etc.)
- Removes the "Connecting to space" Listr task (`getSpace`/`getEnvironment`) from `push-to-space.ts` — the plain client takes `spaceId`/`environmentId` directly as params
- Threads `client`, `spaceId`, and `environmentId` through the call stack wherever entity operations are performed
- Updates `PushToSpaceContext` to carry `{ client, spaceId, environmentId }` instead of a legacy `target` environment object
- Updates all unit tests to mock plain client namespace methods instead of entity instance methods

## Test plan

- [x] All unit tests pass (`npm test`) — 91 passing, yargs tests pre-existing failure (require built `dist/`)
- [x] TypeScript compiles clean (`npx tsc --noEmit`)
- [x] Integration test against a real space

### DIFF
--- a/lib/tasks/get-destination-data.ts
+++ b/lib/tasks/get-destination-data.ts
@@ -1,46 +1,56 @@
 import Promise from 'bluebird'
 
 import { logEmitter } from 'contentful-batch-libs/dist/logging'
-import type { AssetProps, ContentTypeProps, EntryProps, LocaleProps, TagProps, WebhookProps } from 'contentful-management'
+import type { AssetProps, ContentTypeProps, EntryProps, LocaleProps, PlainClientAPI, TagProps, WebhookProps } from 'contentful-management'
 import { OriginalSourceData } from '../types'
 import PQueue from 'p-queue'
 
 const BATCH_CHAR_LIMIT = 1990
 const BATCH_SIZE_LIMIT = 100
 
-const METHODS = {
-  contentTypes: { name: 'content types', method: 'getContentTypes' },
-  locales: { name: 'locales', method: 'getLocales' },
-  entries: { name: 'entries', method: 'getEntries' },
-  assets: { name: 'assets', method: 'getAssets' },
-  tags: { name: 'tags', method: 'getTags' }
-}
-
 type BatchedIdQueryParams = {
   requestQueue: PQueue
-  environment: any
-  type: keyof typeof METHODS
+  client: PlainClientAPI
+  spaceId: string
+  environmentId: string
+  type: 'contentTypes' | 'entries' | 'assets'
   ids: string[]
 }
 
-type BatchedPageQueryParams = Omit<BatchedIdQueryParams, 'ids'>
+type BatchedPageQueryParams = {
+  requestQueue: PQueue
+  client: PlainClientAPI
+  spaceId: string
+  environmentId: string
+  type: 'locales' | 'tags'
+}
 
-async function batchedIdQuery ({ environment, type, ids, requestQueue }: BatchedIdQueryParams) {
-  const method = METHODS[type].method
-  const entityTypeName = METHODS[type].name
+const ENTITY_METHODS = {
+  contentTypes: { name: 'content types', ns: 'contentType' },
+  entries: { name: 'entries', ns: 'entry' },
+  assets: { name: 'assets', ns: 'asset' },
+  locales: { name: 'locales', ns: 'locale' },
+  tags: { name: 'tags', ns: 'tag' },
+} as const
+
+async function batchedIdQuery ({ client, spaceId, environmentId, type, ids, requestQueue }: BatchedIdQueryParams) {
+  const { name, ns } = ENTITY_METHODS[type]
   const batches = getIdBatches(ids)
 
   let totalFetched = 0
 
   const allPendingResponses = batches.map((idBatch) => {
-    // TODO: add batch count to indicate that it's running
     return requestQueue.add(async () => {
-      const response = await environment[method]({
-        'sys.id[in]': idBatch,
-        limit: idBatch.split(',').length
+      const response = await (client[ns] as any).getMany({
+        spaceId,
+        environmentId,
+        query: {
+          'sys.id[in]': idBatch,
+          limit: idBatch.split(',').length
+        }
       })
       totalFetched = totalFetched + response.items.length
-      logEmitter.emit('info', `Fetched ${totalFetched} of ${response.total} ${entityTypeName}`)
+      logEmitter.emit('info', `Fetched ${totalFetched} of ${response.total} ${name}`)
 
       return response.items
     })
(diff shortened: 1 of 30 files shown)


michaelphamcf:
2 things that would help raise confidence before merging:

**1. Add a unit test for the locale `internal_code` fix (commit `d9ab5b0`).**
`test/unit/tasks/push/creation.test.ts` — current tests only assert `client.locale.update` was called, not what fields it was called with. Add a case where `destinationEntity` has `internal_code` set (matches what the plain client's `getMany` actually returns) and assert the update payload doesn't contain it. This is the direct regression test for the 422 bug this PR fixes.

**2. The webhook `createWithId` workaround needs more than a unit test.**
The plain client has no `client.webhook.createWithId()` — `createInDestination()` in `creation.ts` works around this by calling `client.webhook.update()` instead when an id is present. A unit test only proves the code calls what you intended; it can't prove `update()` actually behaves like a create against the real CMA API (different version-conflict semantics). 
- add an integration test case (there's already a `webhooks: []` array in `test/integration/exports/simple/sample-space.json` you can populate, then assert via `client.webhook.getMany()` in `import-lib.test.ts`)

michaelphamcf (dismissed):

lib/tasks/get-destination-data.ts
```diff
@@ -1,7 +1,7 @@
 import Promise from 'bluebird'
```
ethan-ozelius-contentful, line 1:
```

Before: called client.getSpace() → space.getEnvironment() → then asked the environment object for its content types/entries/etc. (dynamic dispatch: environment[methodName]()).
Now: no space/environment objects exist anymore. Just call client.contentType.getMany({spaceId, environmentId, ...}), client.entry.getMany(...), etc. directly.
Why: the plain client has no "environment object" to ask — you always pass spaceId/environmentId as plain arguments to every call. So this file got rewritten to thread those two IDs through instead of holding an environment handle.
```

lib/tasks/push-to-space/push-to-space.ts
ethan-ozelius-contentful, line 1:
```
1. Deleted the "Connecting to space" step. It used to exist only to produce that environment object everything else needed. Not needed anymore → deleted.
2. Every task in this file (editor interfaces, uploads, webhooks, etc.) that used to call a method on an entity/environment object now calls the equivalent client.xxx.yyy(ids, data) form.
3. This is where I found the actual bug, by the way: one leftover task ("Create ExO Folders") still reads ctx.space.sys.organization.sys.id — but ctx.space was only ever set by that now-deleted "Connecting to space" step. Nobody re-wired it, so that value is undefined today. A later commit noticed this crashes and "fixed" it by wrapping it in a try/catch that just logs a warning and moves on — so right now, ExO folder creation silently does nothing instead of throwing. Needs a real fix (fetch org id via client.space.get({spaceId}), which the codebase already does elsewhere), not a bigger try/catch.
```

ethan-ozelius-contentful:
@michaelphamcf 
> 2 things that would help raise confidence before merging:
> 
> **1. Add a unit test for the locale `internal_code` fix (commit `d9ab5b0`).** `test/unit/tasks/push/creation.test.ts` — current tests only assert `client.locale.update` was called, not what fields it was called with. Add a case where `destinationEntity` has `internal_code` set (matches what the plain client's `getMany` actually returns) and assert the update payload doesn't contain it. This is the direct regression test for the 422 bug this PR fixes.
> 
> **2. The webhook `createWithId` workaround needs more than a unit test.** The plain client has no `client.webhook.createWithId()` — `createInDestination()` in `creation.ts` works around this by calling `client.webhook.update()` instead when an id is present. A unit test only proves the code calls what you intended; it can't prove `update()` actually behaves like a create against the real CMA API (different version-conflict semantics).
> 
> * add an integration test case (there's already a `webhooks: []` array in `test/integration/exports/simple/sample-space.json` you can populate, then assert via `client.webhook.getMany()` in `import-lib.test.ts`)

1. Added a case in creation.test.ts ("Strips server-managed fields like internal_code when updating a locale") that sets internal_code on the destinationEntity

2. extended the existing `"It should import a space properly when used as a lib" test` to call client.webhook.getMany() and assert that the webhook was created with the correct id.

lib/tasks/push-to-space/push-to-space.ts
michaelphamcf: I feel #3 is not a blocker for this PR. Log warning and optimistically continue is what the rest of import process does, so it seems to put this in line with the rest of import. 

That said, the suggested fix seems like a small lift to look into implementing in a follow-up PR.

michaelphamcf (approved):

Issue #1671, opened 2026-07-14
[BUG] ESM endpoint cannot be imported under Node.js 22/24


ZachMHamilton:
## Bug Description

The published ESM entry point (dist/index.mjs) cannot be imported under Node.js 22/24. The generated bundle contains CommonJS syntax (module.exports) and invalid lodash named imports, causing runtime failures.

Changing the package exports to point 'import' to dist/index.js works around this issue, indicating the CommonJS build is functional while the ESM build is not.

## Steps to Reproduce

1. Create a Node.js project using Nodejs 22 or 24
2. Install contentful-import@10.0.18
3. Import and use the package as described

## Expected Behavior

Package should import successfully and execute the import using the published ESM build.

## Actual Behavior

ESM build fails during module loading. Initially I encountered import errors with lodash (see error messages below). After correcting those imports locally, another runtime error occurred due to CommonJS 'module.exports' syntax.

## Code Sample

```javascript
import spaceImport from "contentful-import";

await spaceImport({
    // valid import options
})
```

## Environment

- OS: Windows 11
- Package Version: 10.0.18
- Node Version: 24.x
- Package Manager: npm

## Error Messages/Logs

Invalid lodash imports
```
file:.../node_modules/contentful-import/dist/index.mjs:838
import { omit as omit3, defaults } from "lodash/object";
                        ^^^^^^^^
SyntaxError: The requested module 'lodash/object' does not provide an export named 'defaults'
```

CommonJS syntax in ESM build
```
file:.../node_modules/contentful-import/dist/index.mjs:1419
module.exports = runContentfulImport;
^

ReferenceError: module is not defined in ES module scope
```

## Screenshots

N/A

## Additional Context

Changing package.json exports to: 
```
"exports": {
    ....
    "import": "./dist/index.js" // CJS build
}
```

## Possible Solution

Temporary workaround is to change the package.json 'import' export value to point to ./dist/index.js for ease of use for consumers. 

The file lib/index.ts contains both module.exports and export default syntax, citing interop issues with the 'cli-table3' package. This comment is from 3+ years ago, recommendation is to investigate if this is still an issue

ethan-ozelius-contentful:
Hi @ZachMHamilton, thanks for raising this.  Yeah, I noticed this last week as well and we have an internal ticket for it.  I'll try to get a fix for this out by the end of the week.

ethan-ozelius-contentful:
[...]

PR #1672, opened 2026-07-14
fix: repair broken ESM build caused by CJS-only sub-path imports [AIS-207]

ethan-ozelius-contentful:
### Root Cause
The published ESM build (dist/index.mjs) has never worked under Node's ESM loader, for 3 reasons:
1. Lodash/date-fns "sub-path" imports. The old code did things like `import { omit } from 'lodash/object'` and `import format from 'date-fns/format'`. Those sub-paths are internal CJS files inside those packages — fine for require(), but Node's ESM resolver refuses to load them ("does not provide an export named...").
2. contentful-batch-libs has no ESM support at all. It's a CJS-only dependency with no "exports map" (a package.json field that tells Node's ESM resolver how to find sub-paths). The build tool (tsup) normally leaves dependencies external (not bundled), so this unresolvable import got shipped straight into the ESM output and blew up at runtime.
3. Literal `module.exports` in the source. lib/index.ts had both export default runContentfulImport and `module.exports = runContentfulImport` in the same file — a leftover workaround for an old cli-table3 interop issue. module.exports is CJS-only syntax; when that file gets compiled into the ESM build, Node throws ReferenceError: module is not defined.

### Solution:
1. Fixed the imports
  - upgrade lodash to lodash-es
  - upgrade date-fns to a version that supports ESM natively
  - contentful-batch-libs does not have an available ESM build
2. bundle problem dependencies:
  - added tsup.config.ts (previously build config lived in package.json) with `noExternal: ['contentful-batch-libs', 'lodash', 'date-fns']` — this tells tsup to actually inline those packages' code into the output instead of leaving them as external import/require calls the ESM loader can't resolve.
3. remove `module.exports` from anywhere in code.
 

This PR fixes it while preserving full backward compatibility for CJS consumers. Example:

```js
// package.json
{
  "name": "import-esm-cjs-test",
  "version": "1.0.0",
  "scripts": {
    "test:cjs": "node test-cjs.js",
    "test:esm": "node test-esm.mjs"
  },
  "dependencies": {
    "contentful-import": "^10.0.18"
  }
}

// test-cjs.js
const spaceImport = require('contentful-import')

console.log('typeof spaceImport:', typeof spaceImport)
console.log('Is callable:', typeof spaceImport === 'function' ? 'YES ✓' : 'NO ✗ (got ' + JSON.stringify(Object.keys(spaceImport)) + ')')

// test-esm.mjs
import spaceImport from 'contentful-import'

console.log('typeof spaceImport:', typeof spaceImport)
console.log('Is callable:', typeof spaceImport === 'function' ? 'YES ✓' : 'NO ✗ (got ' + JSON.stringify(Object.keys(spaceImport)) + ')')
```

#### Result
```
npm run test:esm

> import-esm-cjs-test@1.0.0 test:esm
> node test-esm.mjs

node:internal/modules/run_main:107
    triggerUncaughtException(
    ^

Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/Users/ethan.ozelius/Development/developer-experience-os/scratch/import-esm-cjs-test/node_modules/date-fns/differenceInSeconds' is not supported resolving ES modules imported from /Users/ethan.ozelius/Development/developer-experience-os/scratch/import-esm-cjs-test/node_modules/contentful-import/dist/index.mjs
Did you mean to import "date-fns/differenceInSeconds/index.js"?
    at finalizeResolution (node:internal/modules/esm/resolve:262:11)
    at moduleResolve (node:internal/modules/esm/resolve:864:10)
    at defaultResolve (node:internal/modules/esm/resolve:990:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:718:20)
    at #resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:735:38)
    at ModuleLoader.resolveSync (node:internal/modules/esm/loader:764:52)
    at #resolve (node:internal/modules/esm/loader:700:17)
    at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:620:35)
    at ModuleJob.syncLink (node:internal/modules/esm/module_job:143:33)
    at ModuleJob.link (node:internal/modules/esm/module_job:228:17) {
  code: 'ERR_UNSUPPORTED_DIR_IMPORT',
  url: 'file:///Users/ethan.ozelius/Development/developer-experience-os/scratch/import-esm-cjs-test/node_modules/date-fns/differenceInSeconds'
}

Node.js v24.13.0
```

**Root cause**

Three issues combined to break the ESM build:

1. lodash and date-fns sub-path imports — imports like lodash/collection, lodash/object, and date-fns/format are CJS-only internal modules. Node's strict ESM resolver rejects them at runtime with SyntaxError: The requested module does not provide an export named .... Fixed by importing from the top-level package entry instead.

5. contentful-batch-libs is CJS-only with no exports map — tsup externalizes all node_modules by default, so import { logEmitter } from 'contentful-batch-libs/dist/logging' was passed straight through into the ESM output. Node's ESM resolver requires an exports map to resolve sub-paths and throws ERR_MODULE_NOT_FOUND. Fixed by bundling contentful-batch-libs (and its transitive CJS deps) directly into the output.
6. module.exports in source — lib/index.ts contained module.exports = runContentfulImport as a workaround for CJS interop. This is invalid in ESM scope and causes `ReferenceError: module is not defined`. Replaced with a build-time footer in tsup.config.ts scoped to the CJS output only.

**Solution**
- Migrate tsup config from package.json to tsup.config.ts to allow per-format build options
- Bundle contentful-batch-libs, lodash, and date-fns into the output (noExternal) so the ESM build has no unresolvable CJS sub-path imports at runtime
- Add createRequire banner to ESM output to give bundled CJS code access to Node built-ins
- Add CJS footer to preserve require('contentful-import') returning the function directly (no .default needed)
- Replace all lodash/* and date-fns/* sub-path imports with top-level imports

**Test plan**
- [x] node test-esm.mjs in a consumer project — typeof spaceImport === 'function'
- [x] node test-cjs.js in a consumer project — typeof spaceImport === 'function'
- [x] npm run test:unit — 90/90 tests pass

**Notes**
Bundling lodash and date-fns increases the ESM artifact size (~415KB vs ~51KB before). This is a known tradeoff given that lodash v4 and date-fns v2 are CJS-only. A future major version should upgrade to lodash-es / date-fns v3+ and drop (potentially) the CJS build entirely.

Closes issue #[1671](https://github.com/contentful/contentful-import/issues/1671)


michaelphamcf:
Hey Ethan, possible regression. The actual contentful-import CLI binary crashes now:
```
TypeError: Cannot read properties of undefined (reading 'config')
    at parseOptions (dist/index.js:11085:29)
```
**Root cause**: the new `tsup.config.ts` footer (`module.exports = module.exports.default || module.exports`) already unwraps `dist/usageParams.js`'s export. `bin/contentful-import` still does `.default` on top of that →
`undefined` → boom.

**Repro**:
```
gh pr checkout 1672
npx tsup
node bin/contentful-import --space-id x --management-token y --content-file test/fixtures/1LargeFile/contentfulExport-01.json
```
Crashes on PR branch, works fine (gets past parseOptions) on main.

**Fix**: drop the `.default` in `bin/contentful-import`:
```
const usageParams = require('../dist/usageParams') // no .default
```

bin/contentful-import
ethan-ozelius-contentful, line 1: 1. Remove `.default` to fix downstream bug in CJS builds.

2. non-functional change from `.then()` to `async/await`

test/unit/contentful-import-bin.test.ts
ethan-ozelius-contentful, line 1: smoke test to ensure that the `bin/contentful-import` executable runs without crashing.  This covers the `.default` [bug](https://github.com/contentful/contentful-import/pull/1672/#issuecomment-4985813792) that was discovered:

michaelphamcf (approved):

PR #1676, opened 2026-07-15
fix: Sort Component & Fragment by references [AIS-239]

ethan-ozelius-contentful:
## Context
(Traditional) Contentful entities can obviously reference each other, for example an entry can reference an asset.  Traditional contentful entities like entry/asset use (regular) links to reference other entities, that are NOT validated at create/update time.  The import just **trusts** that when the referenced entity is needed (delivery for example) it will exists.

ExO Contentful entities use `ResourceLink`s that **DO get validated** at create/update time.  Therefore, if entity `A` references, entity `B`, then `B` must be imported before `A`. To accomplish this, this commit adds a "Kahn" topological sorting algorithm to build a reference graph of all the entities to be imported and in what order. 

## Summary
- **Topological sort for Exo entities** — entity types can reference other entities of the same type (`componentTree` for CTs, `slots` for Fragments). Without sorting, a composite CT can be sent before its leaf dependencies exist, causing `404` errors. Kahn's algorithm is applied to both before sequential import.

### Reference dependencies:
| Entity | Depends On |
|---|---|
| DataAssemblies | _(none)_ |
| ComponentTypes | DataAssemblies, other ComponentTypes |
| Templates | ComponentTypes |
| Fragments | ComponentTypes, DataAssemblies |
| Experiences | Templates, Fragments, DataAssemblies |

- **Documentation** — added `docs/exo-import.md` covering the entity dependency model, why load order matters, why intra-type sorting is required, and why ID preservation is required.

## Root cause

The CMA validates ExO entity references **at create time**, not lazily like standard entry links. Creating a `ComponentType` that references a `DataAssembly` URN before that DA exists → `422`. Creating a composite CT that references another CT by ID before that CT exists → `404`. Both errors were caused by the same underlying issue: `POST`-based creates throwing away the original `sys.id`.

## Test plan

- [x] Import the `nextjs-cko-demo` example space (29 ComponentTypes, 1 Template, 9 DataAssemblies) into a fresh environment and verify all entities are created with their original IDs
- [x] Verify no `404 Not Found` or `422 InvalidDataAssemblyReference` errors in the error log for ComponentType or DataAssembly creates
- [x] Verify composite ComponentTypes (e.g. `metricsSection`, `fourRowShowcase`) are created after their leaf dependencies (`metric`, `section`, `title`, etc.)
- [ ] Re-running import on an already-populated space (update path) should still work correctly

### DIFF
--- a/lib/tasks/push-to-space/push-to-space.ts
+++ b/lib/tasks/push-to-space/push-to-space.ts
@@ -16,6 +16,8 @@ import * as creation from './creation'
 import * as publishing from './publishing'
 import type { DestinationData, TransformedSourceData, Resources, TransformedAsset } from '../../types'
 import { ContentfulEntityError } from '../../utils/errors'
+import sortComponentTypes from '../../utils/sort-component-types'
+import sortFragments from '../../utils/sort-fragments'
 
 const DEFAULT_CONTENT_STRUCTURE = {
   entries: [],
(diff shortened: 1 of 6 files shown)


lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -106,6 +103,10 @@ export default function pushToSpace({
   assetsDirectory,
   requestQueue
 }: PushToSpaceParams) {
+
+  console.log("<><> push-to-space <><> sourceData: ", sourceData)
+  console.log("<><> push-to-space <><> destinationData: ", destinationData)
```
michaelphamcf: remove console logs?

michaelphamcf (changes requested):

lib/tasks/push-to-space/push-to-space.ts
ethan-ozelius-contentful: Good catch, my bad.  I thought I had pushed a change to remove those.

ethan-ozelius-contentful: - [x] remove

package.json
```diff
@@ -100,7 +100,8 @@
   "files": [
     "bin",
     "dist",
-    "example-config.json"
```
ethan-ozelius-contentful, line 103: revert

ethan-ozelius-contentful: - [x] revert

michaelphamcf (approved):

PR #1677, opened 2026-07-16
BackOff - retry DataAssembly writes on stale GraphQL schema after content type publish [AIS-343]

ethan-ozelius-contentful:
## Summary

- Adds exponential backoff retry logic to DataAssembly creates and updates, specifically targeting 422 errors caused by a stale SchemaRegistry response after content types are published in the same import run.

## Background

The `assemblies-management-api` fetches a fresh GraphQL schema from the SchemaRegistry on every DataAssembly write and validates the resolver query and parameter `allowedTypes` against it. When content types are published for the first time in a space (as part of the same `contentful-import` run), the SchemaRegistry has a propagation delay before those types appear in the SDL. This causes valid DataAssembly writes to fail with:

- `GraphQL validation error: Cannot query field "X" on type "Query"`
- `Unknown content type "X"`
- `Pointer path does not exist for '$resolvers/main/X/...'`

There is no bypass or skip flag in the server-side validation layer. This is a confirmed server-side propagation delay (investigated via the `assemblies-management-api` source and confirmed in [#exo-prod-dev-extended-team](https://contentful.slack.com/archives/C0B7THAABCK/p1784233810878919)).

## What changed

Added two helpers in `push-to-space.ts`:

- **`isGraphQLSchemaStaleError`** — detects the specific 422 error signatures caused by a stale schema. All other errors (real validation failures, network errors, etc.) are not retried.
- **`withGraphQLSchemaBackoff`** — retries up to 3 times with delays of 500ms, 1500ms, and 6000ms (~8s total worst case). Logs a warning on each retry. Only wraps DataAssembly writes — all other ExO entity types do not perform GraphQL schema validation.

## Test plan

- [ ] Import into a **fresh environment** (no pre-existing content types) — DataAssembly writes should retry and eventually succeed once the schema propagates
- [ ] Import into a **pre-populated environment** — no retries should trigger, import runs at full speed
- [ ] A genuine validation error (e.g. malformed resolver query) should still fail immediately without retrying

### DIFF
--- a/lib/tasks/push-to-space/push-to-space.ts
+++ b/lib/tasks/push-to-space/push-to-space.ts
@@ -16,8 +16,27 @@ import * as creation from './creation'
 import * as publishing from './publishing'
 import type { DestinationData, TransformedSourceData, Resources, TransformedAsset } from '../../types'
 import { ContentfulEntityError } from '../../utils/errors'
+import { GRAPHQL_SCHEMA_STALE_DELAYS_MS, isGraphQLSchemaStaleError } from '../../utils/graphql-schema-backoff'
 import sortComponentTypes from '../../utils/sort-component-types'
 import sortFragments from '../../utils/sort-fragments'
+
+async function withGraphQLSchemaBackoff<T>(fn: () => Promise<T>): Promise<T> {
+  let lastErr: unknown
+  for (let attempt = 0; attempt <= GRAPHQL_SCHEMA_STALE_DELAYS_MS.length; attempt++) {
+    try {
+      return await fn()
+    } catch (err) {
+      if (!isGraphQLSchemaStaleError(err) || attempt === GRAPHQL_SCHEMA_STALE_DELAYS_MS.length) {
+        throw err
+      }
+      const delay = GRAPHQL_SCHEMA_STALE_DELAYS_MS[attempt]
+      logEmitter.emit('warning', `DataAssembly GraphQL schema not yet current, retrying in ${delay}ms (attempt ${attempt + 1}/${GRAPHQL_SCHEMA_STALE_DELAYS_MS.length})`)
+      await new Promise((resolve) => setTimeout(resolve, delay))
+      lastErr = err
+    }
+  }
+  throw lastErr
+}
 
 const DEFAULT_CONTENT_STRUCTURE = {
   entries: [],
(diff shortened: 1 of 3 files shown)


lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -19,6 +19,36 @@ import { ContentfulEntityError } from '../../utils/errors'
 import sortComponentTypes from '../../utils/sort-component-types'
 import sortFragments from '../../utils/sort-fragments'
 
+const GRAPHQL_SCHEMA_STALE_DELAYS_MS = [500, 1500, 6000]
+
+function isGraphQLSchemaStaleError(err: unknown): boolean {
+  const details = (err as any)?.data?.details?.errors
```
michaelphamcf: Just wanted to double-check this with you. The analysis I ran noted that the actual real 422 stale-schema  error object thrown may not have a `data` field, resulting in `isGraphQLSchemaStaleError` always returning `false`. If I remember correctly, you said it was working so just wanted to make sure this isn't a problem. Maybe adding a test touching `withGraphQLSchemaBackoff` with an actual example of the error would be good? Open to your thoughts.

ethan-ozelius-contentful: You are a legend and 100% correct, good catch.

When a validation error is returned by the API, and handled by contentful-sdk-core's [error-handler](https://github.com/contentful/contentful-sdk-core/blob/master/src/error-handler.ts) only includes `{ name: "ValidationFailed", message: "<JSON string>"   // the whole errorData blob as a JSON string }`.

I ran a full test and confirmed:

```
const parsed = JSON.parse(err.message)
console.log('[graphql-schema-backoff] Object.keys err:', Object.keys(err))
console.log('[graphql-schema-backoff] err.name:', err.name)
console.log('[graphql-schema-backoff] parsed err.message:', JSON.stringify(parsed, null, 2))

// results in
[graphql-schema-backoff] Object.keys err: [ 'name', 'message' ]
[graphql-schema-backoff] err.name: ValidationFailed
[graphql-schema-backoff] parsed err.message: {
"status": 422,
  "statusText": "",
  "message": "Validation error",
  "details": {
    "errors": [
      {
        "name": "unrecognized_keys",
        "path": [
          "sys"
        ],
        "details": "Invalid input at \"sys\""
      }
    ]
  },
  "request": {
    "url": "/spaces/ds7bjzs48gm7/environments/master/data_assemblies/36REZwyrE2fniT4fOrXByP",
    "headers": {
      "Accept": "application/json, text/plain, */*",
      "Content-Type": "application/vnd.contentful.management.v1+json",
      "X-Contentful-User-Agent": "sdk contentful-management-plain.js/0.0.0-determined-by-semantic-release; platform node.js/v24.13.0; os macOS/v24.13.0;",
      "Authorization": "Bearer ...kmjrQ",
      "X-Contentful-Version": "1",
      "User-Agent": "axios/1.18.1",
      "Content-Length": "3387",
      "Accept-Encoding": "gzip, compress, deflate, br"
    },
    "method": "put",
    "payloadData": "{\"sys\":{\"id\":\"36REZwyrE2fniT4fOrXByP\",\"type\":\"DataAssembly\",\"dataTy...
```

So the resolution is the parse just the error.message, since it is a stringified object, and only after that can we interrogate `.data.details.error` etc.  I also created a new type specifically for this error, instead of casting as any.

```ts
type ContentfulValidationError = {
  details?: {
    errors?: Array<{ message?: string }>
  }
}
```

ethan-ozelius-contentful: - [x] `ContentfulValidationError` type
- [x] parse `error.message` before interrogating
- [x] Add `?` and type safety for `error?.message.details.errors[n]`

michaelphamcf (approved):

PR #1682, opened 2026-07-30
Add Taxonomy folders for exo entity folders [AIS-344]

ethan-ozelius-contentful:
## Summary
ExO entities (with the exepction of `dataAssemblies`) can be organized into folders.  Under the hood, these folders are using taxonomy concepts, so those taxonomies that are defined on the exo entity needs to be imported into the destination space.

Screenshot 2026-07-30 at 9 50 00 AM

More context here, in the [exo folders RFC](https://contentful.atlassian.net/wiki/spaces/PROD/pages/6450975708/ExO+Entity+Folders), and this [slack thread](https://contentful.slack.com/archives/C0BAWGDJL9G/p1784657001624019).

## Description
Importing exo folders is a 5 step process:
1. ensure "parent folder groups" concept-schemes exist in the target organization
2. derive the list of "child" folder concepts to create in target space, based on
exo entities.metadata.concepts that match 'contentful-contentful.folder-' prefix
3. create/update the "child" concepts in the targets concepts in the target space
4. link each "child" concept to its associated "parent" folder group
5. rewrite each exo entity's metadata.concepts to point to newly created concept,
instead of concept that is scoped to destination space.

[See more detail in doc writeup](https://github.com/contentful/contentful-import/pull/1682/changes#diff-4a35a58c48d093695c0c6b11c7f391d32aefa85efd695ae071e18d30e99f714f): `docs/exo-import.md`


lib/utils/patch-folder-concepts.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const FOLDER_CONCEPT_PREFIX = 'folder-'
+
+type ExoFolderEntity = ComponentTypeProps | DesignTokenProps | TemplateProps | FragmentProps | ExperienceProps
+
+function collectFolderConceptIds(entities: ExoFolderEntity[]): Set<string> {
+  const ids = new Set<string>()
+  for (const entity of entities) {
+    for (const concept of entity.metadata?.concepts ?? []) {
+      if (concept.sys.id.includes(FOLDER_CONCEPT_PREFIX)) {
```
michaelphamcf: Is it better to narrow this to `startsWith('contentful.folder-')` to be more specific to the concept created for exo? A non-exo concept such as `my-folder-2026` could be picked up when it shouldn't be?

lib/utils/patch-folder-concepts.ts
```diff
@@ -0,0 +1,95 @@
...
+        path: '/metadata/spaces/-',
+        value: { sys: { type: 'Link', linkType: 'Space', id: spaceId } },
+      },
+    ]
+
+    try {
+      await plainClient.concept.patch({ organizationId, conceptId, version: concept.sys.version }, patch)
+      logEmitter.emit('info', `Patched folder concept ${conceptId} to include space ${spaceId}`)
+    } catch (err: any) {
+      logEmitter.emit('warning', `Failed to patch folder concept ${conceptId}: ${err?.message ?? err}`)
```
michaelphamcf: would there be a scenario where the customer is running multiple imports or whatever that would patch the same concept? If this catch an error, does that mean the destination space doesn't get linked to the folder concept, but logs the warning and the customer would have troubleshoot from there?

lib/utils/import-exo-folders.ts
```diff
@@ -0,0 +1,299 @@
...
+      // Non-critical — fall back to the dest concept ID as label
+    }
+
+    // Check whether the destination concept already exists.
+    let existing: any = null
+    try {
+      existing = await plainClient.concept.get({ organizationId, conceptId: destConceptId })
+    } catch (err: any) {
+      if (err?.status !== 404 && err?.sys?.id !== 'NotFound') {
+        // Fail silently as 404 error is expected when the concept doesn't exist yet.
```
michaelphamcf: All errors might fail silently since error from sdk looks like below, and `err.status` and `err.sys.id` seem to be the raw HTTP error response

```{
  [stack]: [Getter/Setter],
  name: 'NotFound',
  message: '{\n' +
    '  "status": 404,\n' +
    '  "statusText": "",\n' +
    '  "message": "The resource could not be found.",\n' +
    '  "details": {},\n' +
    '  "request": {\n' +
    '    "url": "/organizations/<org_id>/taxonomy/concepts/verify-404-shape-does-not-exist-1786049200652",\n' +
    '    "headers": {\n' +
    '      "Accept": "application/json, text/plain, */*",\n' +
    '      "Content-Type": "application/vnd.contentful.management.v1+json",\n' +
    '      "X-Contentful-User-Agent": "sdk contentful-management-plain.js/0.0.0-determined-by-semantic-release; platform node.js/v24.16.0; os macOS/v24.16.0;",\n' +
    '      "Authorization": "Bearer ...",\n' +
    '      "User-Agent": "axios/1.18.1",\n' +
    '      "Accept-Encoding": "gzip, compress, deflate, br"\n' +
    '    },\n' +
    '    "method": "get"\n' +
    '  }\n' +
    '}'
}
```

Potential fix:
```
} catch (err: any) {
  if (err?.name !== 'NotFound') {
    logEmitter.emit('warning', `Could not fetch destination child concept ${destConceptId}: ${err?.message ?? err}`)
  }
  // NotFound falls through silently — existing stays null, proceeds to create
}
```

lib/utils/import-exo-folders.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  // Step 5
+  const allEntities = [
+    ...(sourceEntities.designTokens ?? []),
+    ...(sourceEntities.components ?? []),
+    ...(sourceEntities.experienceTemplates ?? []),
+    ...(sourceEntities.experienceFragments ?? []),
+    ...(sourceEntities.experiences ?? []),
+  ]
+  rewriteEntityFolderConcepts(allEntities, childConceptMap)
```
michaelphamcf, line 297: For Step 5, here's a potential scenario that might be an issue:

**The setup**: A customer is migrating a DesignToken called "Brand Primary Color" from Space A to Space B. In Space A, it lives inside a folder called "Colors." Folders are just taxonomy concepts under the hood, so this token's data has a link to concept ID `contentful.folder-colors-cBllAkZ9`.

**Step by step:**

1. The import computes a new ID for what "Colors" should be called in Space B: `contentful.folder-colors-cBllAkZ9-spaceB`.
2. It tries to create that new concept in Space B — but this one call fails. Doesn't matter why (a rate limit, a permissions hiccup, bad luck) — the point is it fails.
3. The code catches that failure, logs a quiet warning, and keeps going rather than stopping.
4. The next step links `contentful.folder-colors-cBllAkZ9-spaceB` into the org-wide "Colors" folder registry — it doesn't check whether step 2 actually succeeded, so it adds the link anyway. The org-wide registry now claims a folder exists that doesn't.
5. The next step rewrites the "Brand Primary Color" token's data to point at `contentful.folder-colors-cBllAkZ9-spaceB` — again, no check on step 2's outcome.
6. The import finishes normally. "Brand Primary Color" gets created in Space B, carrying a link to a folder concept that was never actually created.

**Why it's a problem:**

- The customer's DesignToken now has a broken reference baked into it — a pointer to nothing.
- That broken reference is also now registered in a folder system that's shared across the entire organization, not just Space B — every other space in that org sees the same broken "Colors" registry entry.
- Nobody is told. There's no error, no failed import, nothing in the summary — just a warning line in a log that's easy to scroll past. The customer finds out only if/when they open Space B, look for the token in the folder tree, and something looks off (or, worse, if the Studio UI can't gracefully handle a folder pointing to nothing and breaks the whole tree view for "Colors" in that space).

The core issue: one failed step gets silently "healed over" by the steps after it, instead of the failure being remembered and acted on.

**Suggested fix:**
1. Step 3 (`createOrPatchChildConcepts`) already knows, per concept, whether `createWithId`/`patch` succeeded or failed — it just needs to report that back instead of only logging it. Change it to return the set of concept IDs that succeeded (or, equivalently, remove failed ones from a working copy of the map).
2. Step 4 (`linkChildConceptsToParentGroups`) should only attempt to link concepts that are in that successful set — right now it links regardless of Step 3's outcome, which is how a nonexistent concept gets registered into the shared org-wide scheme.
3. Step 5 (`rewriteEntityFolderConcepts`) should only rewrite `metadata.concepts` for concepts that made it through Steps 3 and 4. For any concept that failed, strip that link off the entity entirely (leave it unfoldered in the destination) rather than leaving it pointing at either a broken destination ID or a stale source-space ID — both of those would still be dangling.
4. Raise the visibility of the failure. Right now it's a `warning` — easy to miss. Log it as an `error` (so it lands in the error log file) and/or surface a count in the final import summary, so whoever ran the import knows N entities didn't get foldered and can go fix it manually, without the entire import having to abort over it.

lib/utils/patch-folder-concepts.ts
ethan-ozelius-contentful: Good call, looks like claude already picked that up when I did a larger refactor.

`if (concept.sys.id.startsWith(FOLDER_CONCEPT_PREFIX) && ...`

lib/utils/import-exo-folders.ts
ethan-ozelius-contentful: Nice, love this.  will implement.

- [x] implement

lib/utils/patch-folder-concepts.ts
ethan-ozelius-contentful: Discussed with @michaelphamcf on zoom, the scenario of running multiple imports is definitely valid, and should be handled by the current code.  As to running the multiple patches on the same child concept, step 2 makes a unique list, so we should be good there.  Lastly running multiple imports at the same time concurrently, there's no way for this client to be aware of that, or mediate it.  That would be an API level error handling situation.

Lastly, lastly, yes if we do catch an error here, it would be on the customer to try to debug a little bit here to figure out what went wrong, since there are no specific error cases that I can think of to handle individually.

lib/utils/import-exo-folders.ts
michaelphamcf: Reproduced this scenario.
 
- Confirmed Step 3 (`createOrPatchChildConcepts`) produces 422, swallowed and logged as warning
- Confirmed Step 4 (`linkChildConceptsToParentGroups`) 422s, so this claim is not true: `adds the link anyway. The org-wide registry now claims a folder exists that doesn't.`
- Step 5, confirmed that the design token is still visible in the UI, but isn't grouped in a folder.
  - Users can move it to the correct folder manually.

michaelphamcf (approved):

PR #1683, opened 2026-07-30
feat: maintain publish state for ExO entities during import [AIS-341]

michaelphamcf:
[AIS-341](https://contentful.atlassian.net/browse/AIS-341)

## Summary
- Adds a "Publishing X" Listr task after each ExO entity's "Importing X" task (DataAssembly, ComponentType, Template, Fragment, Experience), gated on `sys.publishedVersion` in the source and respecting `skipContentPublishing` — matching the existing behavior for entries/assets. Previously, all ExO entities were left in draft regardless of their source publish state.
- ComponentType and Fragment publish sequentially in dependency order (reusing the existing `sortComponentTypes`/`sortFragments` topological sort from creation) since their publish-time validation resolves nested references against the *published* state — a parent can't publish before its nested children are published. DataAssembly/Template/Experience publish in parallel (no intra-type dependency).
- DesignToken is intentionally excluded — confirmed via staging that it has no draft/published state and auto-publishes on every `upsert`.

## Notes for reviewers
- **Archive parity is out of scope.** The ticket's background text mentions archived entities should be "similarly re-archived," but neither the CMA SDK nor the upstream `assemblies` API has a base-entity archive/unarchive endpoint for any of the 5 ExO types (only Experience/Fragment optimization-variant sub-resources have archive, which is unrelated). Posted details as a comment on [AIS-341](https://contentful.atlassian.net/browse/AIS-341).
- **This only maintains the binary PUBLISHED/DRAFT split, not the CHANGED state.** If a source entity was published but has since been edited without republishing (CHANGED — `publishedVersion < version`), this PR's gate (`publishedVersion` truthy) still publishes it, collapsing CHANGED into PUBLISHED and pushing live unsaved edits. This is the exact same pre-existing behavior as entries/assets today — not a regression, but not a fix either. A real fix is a bigger lift than it sounds: for entries, `Entry.getPublished` already exists in the CMA SDK, so it's "just" an export-schema + two-step-import-write change. For ExO, it's currently **not possible** for 4 of the 5 types — `ComponentType`, `Template`, `Fragment`, and `Experience` have no `getPublished`/`getManyPublished` method in the CMA SDK at all (only `DataAssembly` does), so there's no way to even read a "last published" snapshot separate from the current draft yet. Fixing this for ExO needs new upstream `assemblies` API + CMA SDK work before `contentful-export`/`contentful-import` could build on top of it. Tracked as a separate follow-up per [AIS-101](https://contentful.atlassian.net/browse/AIS-101)'s investigation, not blocking this PR.
- **Integration test AC is covered by [AIS-134](https://contentful.atlassian.net/browse/AIS-134)**, not this PR — its test-case list already includes "Publish state is preserved: entities that were published in source are published after import," and it requires new CI infra (`IMPORT_SPACE_ID_EXO` secret) that doesn't exist yet. Verified this PR's behavior manually against real staging instead (see test plan).
- Locale-aware publish state for Experience is also out of scope per [AIS-101](https://contentful.atlassian.net/browse/AIS-101)'s investigation — tracked as a separate follow-up ticket, not blocking this PR.

## Test plan
- [x] 27 new/updated unit tests in `push-to-space-exo.test.ts` covering all 5 types: publish-when-source-published, skip-when-source-draft, respects `skipContentPublishing`, log-and-continue on publish failure, no-op when the entity type isn't in the payload
- [x] `tsc --noEmit` clean on both main and test configs
- [x] No new lint errors
- [x] Manual staging verification against `3fyh0othob6d`: seeded 5 published + 5 draft ExO entity pairs (all types), exported and imported into a fresh environment — all published-in-source entities landed published, all draft-in-source stayed draft
- [x] Re-verified across a second export/import hop (round-trip stability) and again across a genuinely separate destination space (cross-space, no folder-concept entities involved) — same correct result both times
- [x] Verified `skipContentPublishing: true` suppresses all 5 new publish steps and leaves every entity in draft, including ones published in the source

[AIS-341]: https://contentful.atlassian.net/browse/AIS-341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AIS-341]: https://contentful.atlassian.net/browse/AIS-341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/tasks/push-to-space/push-to-space.ts
+++ b/lib/tasks/push-to-space/push-to-space.ts
@@ -5,7 +5,10 @@ import { logEmitter } from 'contentful-batch-libs/dist/logging'
 import { wrapTask } from 'contentful-batch-libs/dist/listr'
 import {
   ComponentTypeProps,
+  DataAssemblyProps,
+  ExperienceProps,
   FragmentProps,
+  TemplateProps,
   UpsertComponentTypeProps,
   UpsertTemplateProps,
   UpsertFragmentProps,
@@ -448,14 +451,14 @@ export default function pushToSpace({
     },
     {
       title: 'Publishing Data Assemblies',
-      task: wrapTask(async (ctx) => {
+      task: wrapTask(async (ctx: { data: { dataAssemblies: DataAssemblyProps[], publishedDataAssemblies: DataAssemblyProps[] } }) => {
         const entitiesToPublish = filterExoEntitiesToPublish(ctx.data.dataAssemblies, sourceData.dataAssemblies || [])
         const results = await Promise.all(entitiesToPublish.map((entity) =>
-          publishExoEntity('DataAssembly', entity, () => plainClient.dataAssembly.publish(
+          publishExoEntity<DataAssemblyProps>('DataAssembly', entity, () => plainClient.dataAssembly.publish(
             { spaceId, environmentId, dataAssemblyId: entity.sys.id, version: entity.sys.version }
           ))
         ))
-        ctx.data.publishedDataAssemblies = results.filter(Boolean)
+        ctx.data.publishedDataAssemblies = results.filter((entity): entity is DataAssemblyProps => entity !== null)
       }),
       skip: () => !includeExperienceOrchestration || skipContentPublishing || !(sourceData.dataAssemblies || []).length
     },
@@ -559,14 +562,14 @@ export default function pushToSpace({
     },
     {
       title: 'Publishing Templates',
-      task: wrapTask(async (ctx) => {
+      task: wrapTask(async (ctx: { data: { templates: TemplateProps[], publishedTemplates: TemplateProps[] } }) => {
         const entitiesToPublish = filterExoEntitiesToPublish(ctx.data.templates, sourceData.templates || [])
         const results = await Promise.all(entitiesToPublish.map((entity) =>
-          publishExoEntity('Template', entity, () => plainClient.template.publish(
+          publishExoEntity<TemplateProps>('Template', entity, () => plainClient.template.publish(
             { spaceId, environmentId, templateId: entity.sys.id, version: entity.sys.version }
           ))
         ))
-        ctx.data.publishedTemplates = results.filter(Boolean)
+        ctx.data.publishedTemplates = results.filter((entity): entity is TemplateProps => entity !== null)
       }),
       skip: () => !includeExperienceOrchestration || skipContentPublishing || !(sourceData.templates || []).length
     },
@@ -643,14 +646,14 @@ export default function pushToSpace({
     },
     {
       title: 'Publishing Experiences',
-      task: wrapTask(async (ctx) => {
+      task: wrapTask(async (ctx: { data: { experiences: ExperienceProps[], publishedExperiences: ExperienceProps[] } }) => {
         const entitiesToPublish = filterExoEntitiesToPublish(ctx.data.experiences, sourceData.experiences || [])
         const results = await Promise.all(entitiesToPublish.map((entity) =>
-          publishExoEntity('Experience', entity, () => plainClient.experience.publish(
+          publishExoEntity<ExperienceProps>('Experience', entity, () => plainClient.experience.publish(
             { spaceId, environmentId, experienceId: entity.sys.id, version: entity.sys.version }
           ))
         ))
-        ctx.data.publishedExperiences = results.filter(Boolean)
+        ctx.data.publishedExperiences = results.filter((entity): entity is ExperienceProps => entity !== null)
       }),
       skip: () => !includeExperienceOrchestration || skipContentPublishing || !(sourceData.experiences || []).length
     }
--- a/lib/tasks/push-to-space/push-to-space.ts
+++ b/lib/tasks/push-to-space/push-to-space.ts
@@ -4,6 +4,8 @@ import verboseRenderer from 'listr-verbose-renderer'
 import { logEmitter } from 'contentful-batch-libs/dist/logging'
 import { wrapTask } from 'contentful-batch-libs/dist/listr'
 import {
+  ComponentTypeProps,
+  FragmentProps,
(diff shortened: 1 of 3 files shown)


lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -607,3 +683,28 @@ function publishEntities({ entities, sourceEntities, requestQueue }) {
 
   return publishing.publishEntities({ entities: entitiesToPublish, requestQueue })
 }
+
+// Find all ExO entities in source content which are published, and filter the
```
ethan-ozelius-contentful: nit: let's move this helper to it's own util.

lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -607,3 +683,28 @@ function publishEntities({ entities, sourceEntities, requestQueue }) {
...
+// created/upserted destination entities down to just those. Mirrors publishEntities
+// above, but ExO source entities aren't wrapped in { original, transformed } and are
+// plain JSON rather than SDK-wrapped instances, so ExO entities get their own gate here.
+function filterExoEntitiesToPublish<T extends { sys: { id: string; version: number } }>(
```
ethan-ozelius-contentful: nit: Instead of using `T`, can we just reuse `ComponentTypeProps`, `TemplateProps`, `FragmentProps` etc. 

Kinda seems like overkill, so happy to leave as is also, I'd defer to PR author.

lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -607,3 +683,28 @@ function publishEntities({ entities, sourceEntities, requestQueue }) {
...
+  entities: T[],
+  sourceEntities: Array<{ sys: { id: string; publishedVersion?: number } }>
```
ethan-ozelius-contentful: Similarly, we can reuse the `TransformedSourceData` type here to type sourceEntities instead of creating a new type with just the props we need.

lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -549,6 +594,24 @@ export default function pushToSpace({
...
       skip: () => !includeExperienceOrchestration || !(sourceData.fragments || []).length
     },
+    {
+      title: 'Publishing Fragments',
+      task: wrapTask(async (ctx) => {
+        const entitiesToPublish = filterExoEntitiesToPublish(ctx.data.fragments, sourceData.fragments || [])
+        // Sorted and sequential, same reasoning as Publishing Component Types: a Fragment
+        // can reference other Fragments in its slots, resolved against published state.
+        const sorted = sortFragments(entitiesToPublish)
+        const results: any[] = []
```
ethan-ozelius-contentful: replace `any[]` with `FragmentProps[]`?

test/unit/tasks/push/push-to-space-exo.test.ts
```diff
@@ -126,6 +143,79 @@ describe('Importing Component Types', () => {
   })
 })
 
+describe('Publishing Component Types', () => {
+  const publishedEntity: any = { sys: { id: 'ct-1', type: 'ComponentType', version: 3, publishedVersion: 2 }, name: 'Hero' }
```
ethan-ozelius-contentful: nit of the year :) but I'm going to raise it because we can hopefully fix this soon: this object really represents an entity that has been published, but currently has unpublished changes.  

Maybe update version and publish version to match for now?

ethan-ozelius-contentful (approved):

lib/tasks/push-to-space/push-to-space.ts
michaelphamcf: Done — moved to lib/utils/publish-exo-entities.ts, matching the sort-component-types.ts/sort-fragments.ts convention.

lib/tasks/push-to-space/push-to-space.ts
michaelphamcf: Tried the union approach — it needs casts at the ComponentType/Fragment sites, and doesn't actually buy safety at the other three (Template/DataAssembly/Experience), since ctx is typed any there and nothing gets checked either way.

Fixed that properly in 4b96c9c instead: added explicit ctx types to those 3 tasks and pinned the generic at each call site. That surfaced a real gap — the generic wasn't resolving to a concrete type, so filter(Boolean) was silently producing an unchecked array. Fixed with a type-guard filter.

Net: kept the generic version, but now all 5 Publishing tasks are actually type-checked instead of 2 of 5. No behavior change, tests still pass.

lib/tasks/push-to-space/push-to-space.ts
michaelphamcf: Done — sourceEntities is now typed against the ExO entity prop types directly (ComponentTypeProps | TemplateProps | FragmentProps | DataAssemblyProps | ExperienceProps) instead of an inline shape. Didn't use TransformedSourceData directly since its ExO fields are just bare arrays of these props types (unlike entries/assets which are wrapped in EntityTransformed), so referencing the props types directly was more precise.

lib/tasks/push-to-space/push-to-space.ts
michaelphamcf: Done — also fixed the sibling any[] in the Publishing Component Types block for consistency (ComponentTypeProps[]).

test/unit/tasks/push/push-to-space-exo.test.ts
michaelphamcf: Good catch — fixed here (and the same mismatch in the Template block). Agreed this test should exercise a clean PUBLISHED entity, not accidentally represent CHANGED.

PR #1684, opened 2026-07-30
fix: correct ExO entity UPDATE payloads for DataAssembly, Fragment, Experience [AIS-341]

michaelphamcf:
[AIS-341](https://contentful.atlassian.net/browse/AIS-341)

## Summary
Two pre-existing bugs found while doing staging verification for AIS-341 (#1683). Both only reproduce when re-importing into a space that already has these entities — the CREATE path was unaffected, which is why they went unnoticed until now.

- **DataAssembly UPDATE** spread the entire source `sys` object (`{ ...entity.sys, version: existing.sys.version }`), including server-managed fields like `publishedVersion`, `createdAt`, etc. The API rejects that with `422 Invalid input at "sys"` on every update. Fixed to send only the fields the API accepts, matching the shape already used on create.
- **Fragment/Experience UPDATE** included `componentType`/`template` in the payload, but those fields are immutable after creation — the API rejects with `400 componentType/template is immutable after creation` on every update. Fixed by omitting them from the update payload; the link is preserved server-side from creation and doesn't need to be resent.

## Test plan
- [x] Unit tests updated: strengthened DataAssembly UPDATE test with realistic source `sys` (previously too minimal to catch the bug), added assertion that server-managed fields are stripped; Fragment/Experience UPDATE tests now assert `componentType`/`template` are omitted from the payload rather than asserting the old (buggy) behavior
- [x] `tsc --noEmit` clean
- [x] No new lint errors beyond the pre-existing baseline on `exo`
- [x] Verified against real staging: re-imported into a space already containing DataAssembly/Fragment/Experience entities from a prior import — all updates now succeed with zero errors, and confirmed the immutable `componentType`/`template` links remain correctly intact post-update

[AIS-341]: https://contentful.atlassian.net/browse/AIS-341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/tasks/push-to-space/push-to-space.ts
+++ b/lib/tasks/push-to-space/push-to-space.ts
@@ -419,7 +419,7 @@ export default function pushToSpace({
             const existing = destinationDataById.dataAssemblies?.get(entity.sys.id)
             let result
             if (existing) {
-              const payload: UpdateDataAssemblyProps = { ...entity, sys: { ...entity.sys, version: existing.sys.version } }
+              const payload: UpdateDataAssemblyProps = { ...omitSys(entity), sys: { id: entity.sys.id, type: 'DataAssembly', dataType: entity.sys.dataType, ...(entity.sys.variant ? { variant: entity.sys.variant } : {}), version: existing.sys.version } }
               result = await withGraphQLSchemaBackoff(() => plainClient.dataAssembly.update(
                 { spaceId, environmentId, dataAssemblyId: entity.sys.id },
                 payload
@@ -531,7 +531,8 @@ export default function pushToSpace({
           try {
             const existing = destinationDataById.fragments?.get(entity.sys.id)
             if (existing) {
-              const payload: UpsertFragmentProps = { ...entity, componentType: entity.sys.componentType, sys: { id: entity.sys.id, type: 'Fragment', version: existing.sys.version } }
+              // componentType is immutable after creation; omit it from the update payload
+              const payload: UpsertFragmentProps = { ...entity, sys: { id: entity.sys.id, type: 'Fragment', version: existing.sys.version } }
               const result = await plainClient.fragment.upsert({ spaceId, environmentId, fragmentId: entity.sys.id }, payload)
               logEmitter.emit('info', `UPDATE Fragment ${entity.sys.id}`)
               results.push(result)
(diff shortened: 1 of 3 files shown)


lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -419,7 +419,7 @@ export default function pushToSpace({
             const existing = destinationDataById.dataAssemblies?.get(entity.sys.id)
             let result
             if (existing) {
-              const payload: UpdateDataAssemblyProps = { ...entity, sys: { ...entity.sys, version: existing.sys.version } }
+              const payload: UpdateDataAssemblyProps = { ...omitSys(entity), sys: { id: entity.sys.id, type: 'DataAssembly', dataType: entity.sys.dataType, ...(entity.sys.variant ? { variant: entity.sys.variant } : {}), version: existing.sys.version } }
```
ethan-ozelius-contentful: is variant strictly required?  I had variants/releases in the follow up epic.

lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -556,7 +557,8 @@ export default function pushToSpace({
           try {
             const existing = destinationDataById.experiences?.get(entity.sys.id)
             if (existing) {
-              const payload: UpsertExperienceProps = { ...entity, template: entity.sys.template, sys: { id: entity.sys.id, type: 'Experience', version: existing.sys.version } }
+              // template is immutable after creation; omit it from the update payload
```
ethan-ozelius-contentful: I'm confused by this comment, this step is for creating an experience...

Maybe instead of saying "immutable", (which makes me think of shouting online matches in 2018 about the merits of functional vs OOP programming, and everyone's gonna be using redux in 3 months) we say: "Once an experience is created, it's template cannot be changed/updated to a different template".

lib/tasks/push-to-space/push-to-space.ts
ethan-ozelius-contentful: Out of scope for this PR, but this `omitSys()` and then immediately re-add is not the most readable code.

ethan-ozelius-contentful (approved):

lib/tasks/push-to-space/push-to-space.ts
michaelphamcf: `sys.variant` already exists in `CreateDataAssemblyProps` and `UpdateDataAssemblyProps` in the installed SDK (`contentful-management@12.12.0`). The CREATE branch (a few lines below, pre-refactor) was already conditionally including it before this PR. We need to keep including it on UPDATE too, otherwise re-importing a DataAssembly that already has a `variant` set would silently drop it.

Also took your suggestion on readability nit: extracted a shared `buildDataAssemblySys()` helper (now in `lib/utils/exo-entity-payloads.ts`) that both CREATE and UPDATE call, so there's one visible `variant`-handling path instead of two duplicated inline ones.

ethan-ozelius-contentful: gotcha, makes sense.  I did not realize variant was already included in the CREATE call.  And obviously it's in the plan to add it, I was just wondering if we were introducing scope/complexity before it was required.  And since this is a dev branch anyway, we're pretty free to do whatever :)

Thanks for adding that context.

PR #1687, opened 2026-07-31
fix: log and rethrow when EXO topological sort throws [AIS-345]

michaelphamcf:
[AIS-345](https://contentful.atlassian.net/browse/AIS-345)

## Summary
- `sortComponentTypes`/`sortFragments` run outside the per-entity try/catch in `push-to-space.ts`, directly inside `wrapTask`. An uncaught throw there (e.g. malformed `componentTree`/`slots` data) bubbles through `wrapTask` and aborts the run without ever reaching the `logEmitter`-driven report.
- Added a `sortOrReport<T>(sortFn)` helper: runs the sort, and on throw, emits via `logEmitter` before rethrowing — preserves today's abort-on-failure behavior (no change to the happy path) while ensuring the failure is visible in the final report.
- Wrapped all 4 call sites: `sortComponentTypes` (Importing + Publishing Component Types), `sortFragments` (Importing + Publishing Fragments).

## Test plan
- [x] New test file `push-to-space-exo-sort-errors.test.ts` mocks the sort utilities to throw, asserts the run still rejects (behavior unchanged) but `logEmitter.emit('error', ...)` fires first with the underlying error
- [x] Full unit suite passes (160/160)
- [x] Lint has no new errors
- [x] Verified real before/after behavior via a driven demo script — before, the error crashed the process with `(no errors captured)` on `logEmitter`, invisible to the JSON error-log file; after, the same abort now captures the full error + stack trace in the report first

[AIS-345]: https://contentful.atlassian.net/browse/AIS-345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/tasks/push-to-space/push-to-space.ts
+++ b/lib/tasks/push-to-space/push-to-space.ts
@@ -28,6 +28,20 @@ import sortComponentTypes from '../../utils/sort-component-types'
 import sortFragments from '../../utils/sort-fragments'
 import { filterExoEntitiesToPublish, publishExoEntity } from '../../utils/publish-exo-entities'
 
+// sortComponentTypes/sortFragments run outside the per-entity try/catch below, directly
+// inside wrapTask. An uncaught throw here (e.g. malformed componentTree/slots data) would
+// bubble through wrapTask and abort the run without ever reaching the logEmitter-driven
+// report. This wrapper logs the failure before rethrowing, preserving today's abort-on-failure
+// behavior while ensuring the error is visible in the final report.
+function sortOrReport<T>(sortFn: () => T[]): T[] {
+  try {
+    return sortFn()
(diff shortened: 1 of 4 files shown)


michaelphamcf:
Follow-up: extracted `sortOrReport` into `lib/utils/sort-or-report.ts` with its own unit test, to match the convention every other helper in `push-to-space.ts` already follows (`sortComponentTypes`, `sortFragments`, `publishExoEntity`, etc. all live in `lib/utils/`). It was previously defined inline in `push-to-space.ts`.

ethan-ozelius-contentful (approved):

PR #1693, opened 2026-08-04
fix: warn and continue when destination space lacks ExO entitlement [AIS-141]

michaelphamcf:
[AIS-141](https://contentful.atlassian.net/browse/AIS-141)

## Summary

`getDestinationData()` had no error handling around its 6 ExO entity fetches (Design Tokens, Components, Experience Templates, Experience Fragments, Data Assemblies, Experiences). A destination space without the `exoM1` entitlement 403s on ExO endpoints (`AccessDenied: exoM1 entitlement required`), which propagated straight out of `getDestinationData()` and aborted the entire import before a single entry, asset, or content type was written — even though none of that content has anything to do with ExO. Given both CLIs default `includeExperienceOrchestration: true`, this was the out-of-the-box experience for any space without ExO enabled.

Three layers fix this:

- **Content gate** — each of the 6 ExO fetches now only fires if the source content actually contains that entity type (mirrors the existing `entries`/`assets`/`locales` pattern). An import with zero ExO entities never probes the destination at all, so there's no entitlement question to even ask for the common case.
- **Proactive entitlement check** (`spaceHasExoM1Entitlement`) — before attempting the 5 exoM1-gated types (Design Tokens, Components, Experience Templates, Experience Fragments, Experiences), checks the destination org's public entitlement set once via `GET /organizations/{orgId}/organization_entitlement_set`. This mirrors `contentful-mcp-server`'s `hasExoM1Entitlement` helper (AIS-191), adapted to check the one org we already know instead of every org the token can see. A confirmed-missing entitlement short-circuits with **one** clear message instead of 5 separate per-type 403s. `dataAssemblies` is deliberately excluded — confirmed live it isn't actually gated by `exoM1` the same way as the other 5 — so it always goes through its own independent fetch-and-catch. If the check itself can't complete (network error, space lookup failure), it fails open to the existing reactive per-type catch rather than treating "inconclusive" as "not entitled."
- **Severity split** in the reactive catch (`cursorPaginatedQueryOrWarn`) — the confirmed "ExO not enabled" case still warns and continues, so it doesn't affect overall run success (the common, expected case once gated on real source content). Any other error on these fetches (bad token, a real outage, etc.) now logs as an `error`, so the overall import correctly reports failure instead of silently succeeding despite dropped ExO content.

## Live verification

Ran the built CLI against two real spaces — one confirmed to lack `exoM1`, one confirmed to have it (checked directly against the entitlement endpoint, not inferred):

| Scenario | Destination | Result |
|---|---|---|
| No ExO content in source | lacks `exoM1` | Exit 0, zero ExO probes at all |
| ExO content (incl. Data Assemblies) | lacks `exoM1` | One combined error for the 5 gated types; Data Assemblies still fetched independently and succeeded; exit 1 |
| ExO content in source | has `exoM1` | Normal fetch, no entitlement error, exit 0 |
| ExO content in source, one type forced to a genuine (real, unmocked) 500 | has `exoM1` | Correctly classified as non-entitlement, logged as an error, did not abort the run |
| No ExO content in source | has `exoM1` | Exit 0, zero ExO probes — proves the content gate wins regardless of destination capability |

## Test plan
- [x] `test/unit/tasks/get-destination-data-exo.test.ts` (18 tests) and new `test/unit/utils/publish-exo-entities.test.ts` (7 tests for `spaceHasExoM1Entitlement`) — cover the content gate, the proactive-check composition (definitively missing / present / inconclusive), the severity split, and that `dataAssemblies` always fetches independently.
- [x] Full unit suite passing, no regressions. `tsc --noEmit` clean. Lint diff-clean against this file's pre-existing baseline.
- [x] Live-verified all 5 scenarios above against real spaces (table above).

[AIS-141]: https://contentful.atlassian.net/browse/AIS-141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/tasks/get-destination-data.ts
+++ b/lib/tasks/get-destination-data.ts
@@ -3,6 +3,7 @@ import Promise from 'bluebird'
 import { logEmitter } from 'contentful-batch-libs/dist/logging'
 import type { AssetProps, ComponentProps, ContentTypeProps, DataAssemblyProps, DesignTokenProps, EntryProps, ExperienceProps, ExperienceFragmentProps, LocaleProps, TagProps, ExperienceTemplateProps, WebhookProps } from 'contentful-management'
 import { OriginalSourceData } from '../types'
+import { isExoEntitlementError, spaceHasExoM1Entitlement } from '../utils/publish-exo-entities'
 import PQueue from 'p-queue'
 
 const BATCH_CHAR_LIMIT = 1990
@@ -170,7 +171,11 @@ async function cursorPaginatedQueryOrWarn (params: CursorPaginatedQueryParams): 
     return await cursorPaginatedQuery(params)
   } catch (err) {
     const { name: entityTypeName } = CURSOR_QUERY_METHODS[params.type]
-    logEmitter.emit('warning', `Skipping ${entityTypeName} import: ${err instanceof Error ? err.message : err}`)
+    if (isExoEntitlementError(err)) {
+      logEmitter.emit('error', new Error(`Skipping ${entityTypeName} import: Experience Orchestration (ExO) is not enabled for this space`))
+    } else {
+      logEmitter.emit('error', err instanceof Error ? err : new Error(String(err)))
+    }
     return []
   }
 }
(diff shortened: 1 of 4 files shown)


lib/tasks/get-destination-data.ts
```diff
@@ -161,6 +161,20 @@ async function cursorPaginatedQuery ({ plainClient, spaceId, environmentId, type
...
+// A destination space without the ExO entitlement 403s on every ExO endpoint. That's a normal,
+// expected outcome (most spaces don't have ExO enabled) rather than a fatal error, so it's
+// handled the same way the `tags` fetch above handles spaces without Tags access: warn and
+// fall back to an empty array instead of failing the whole destination-data fetch.
+async function cursorPaginatedQueryOrWarn (params: CursorPaginatedQueryParams): Promise<any[]> {
+  try {
+    return await cursorPaginatedQuery(params)
+  } catch (err) {
+    const { name: entityTypeName } = CURSOR_QUERY_METHODS[params.type]
+    logEmitter.emit('warning', `Skipping ${entityTypeName} import: ${err instanceof Error ? err.message : err}`)
```
ethan-ozelius-contentful: 1. I wonder if we could improve this error messaging a little to be more human readable.  Something like:

`403 authorized access to /experiences endpoint.  Is Experience Orchestration authorized and enabled for this space?`.

2. could check specifically for 403 vs 404 vs 400 errors, and adjust the message accordingly?

3. If we in fact encounter a 403, can we set the `includeExperienceOrchestration` to false?  This might be a bad idea, because we'd be mutating an argument that gets passed in, with very little visablity.  On the plus side though, it would skip all subsequent exo operations, specifically in `push-to-space.`

ethan-ozelius-contentful: Discussed with @michaelphamcf,

1. would be nice to have, but we need to zero in on what the exact error code/message would be.  It is still a little ambiguous of whether it would be a 403/401/404.  We can either just ask what the message would be from the entitlements team, or set up a personal account/org that definitely does not have the EXO enablement.

2. see 1.

3. This is not something we're going to pursue right now.  The current pattern is to optimistically continue with the import, even if we have a pretty good idea that something will fail.  And the code that would be needed to support skipping subsequent exo calls would be hacky.

ethan-ozelius-contentful (approved):

lib/tasks/get-destination-data.ts
michaelphamcf: - Added a couple of exo gate checks to make sure that org has exo entitlement and has exo content.
- Kept the zero-in on `exoM1` error message as a fallback in case entitlement request fails.
- Other messages log as errors
- Optimistically continue with import if exo content import has errors or lack entitlement.

PR #1694, opened 2026-08-10
test: add integration tests for ExO entity import [AIS-134]

michaelphamcf:
[AIS-134](https://contentful.atlassian.net/browse/AIS-134)

## Summary
- Adds `test/integration/import-exo.test.ts` covering all 5 ACs from AIS-134: creates all 6 ExO entity types, preserves publish state from source, keeps `$self`-relative ResourceLink URNs valid across a cross-space import, confirms `includeExperienceOrchestration` gating (unset → skips ExO tasks even with ExO data present - the ticket named a `skipExperienceOrchestration` flag that doesn't exist in the code), and imports a legacy export with no ExO arrays cleanly.
- Also covers the re-import (update) path, beyond the literal ACs: re-imports the same fixture a second time, then unpublishes one entity in a follow-up import and confirms it propagates to the destination. This is where two real bugs previously surfaced against the live API (immutable `component`/`experienceTemplate` on UPDATE, unpublish never propagating) - both fixed already, but only guarded by mocked unit tests until now.
- Fully self-contained, same pattern as the existing `import-lib.test.ts`: each describe block creates its own throwaway space via `createSpace()` and deletes it in `afterAll` - no new env var, no pre-provisioned space, no GH secret needed.

## Test plan
- [x] `npm run lint` - clean
- [x] `npm run test:unit` - passing, unaffected by this change
- [x] Ran `test/integration/import-exo.test.ts` for real against a live, exoM1-entitled space - all 10 tests pass, entities created/cleaned up correctly, cross-references resolved, skip-flag behavior confirmed, re-import/unpublish-propagation confirmed

[AIS-134]: https://contentful.atlassian.net/browse/AIS-134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/integration/import-exo.test.ts
+++ b/test/integration/import-exo.test.ts
@@ -1,0 +1,192 @@
+import { createClient } from 'contentful-management'
+
+import runContentfulImport from '../../dist/index'
+import { buildExoContent, buildUnpublishedComponentContent, EXO_FIXTURE_IDS } from './utils/exo.utils'
+
+const managementToken = process.env.MANAGEMENT_TOKEN as string
+const orgId = process.env.ORG_ID as string
+const environmentId = 'master'
+
+jest.setTimeout(2 * 60 * 1000) // 2min timeout - covers space create/delete + 6 ExO entity types + publish
+
+// Each describe block below creates its own throwaway space (same pattern as
+// import-lib.test.ts) and deletes it in afterAll.
+
+// Covers: "Import of an ExO export file creates all 6 entity types in destination",
+// "URN rewriting: cross-space import produces valid URNs pointing to destination space/env",
+// and "Publish state is preserved" - one import run, multiple assertions, since all three
+// ACs are properties of the same operation rather than independent scenarios.
+describe('Importing an ExO export with all 6 entity types', () => {
+  let spaceId: string
+  let plainClient
+
+  beforeAll(async () => {
+    const legacyClient = createClient({ accessToken: managementToken }, { type: 'legacy' })
+    const space = await legacyClient.createSpace({ name: 'IMPORT [AUTO] TOOL EXO TMP' }, orgId)
+    spaceId = space.sys.id
+    plainClient = createClient({ accessToken: managementToken })
+
+    await runContentfulImport({
+      spaceId,
+      environmentId,
+      managementToken,
+      content: buildExoContent(EXO_FIXTURE_IDS),
+      includeExperienceOrchestration: true,
+      useVerboseRenderer: true
+    })
+  })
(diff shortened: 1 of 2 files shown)


test/integration/import-exo.test.ts
```diff
@@ -0,0 +1,192 @@
...
+    expect(experience.sys.publishedVersion).toBeDefined()
+
+    expect(experience.sys.experienceTemplate.sys.urn).toBe(
+      `crn:contentful:::experience:spaces/$self/environments/$self/experienceTemplates/${EXO_FIXTURE_IDS.experienceTemplateId}`
+    )
+    const referencedTemplate = await plainClient.experienceTemplate.get({ spaceId, environmentId, experienceTemplateId: EXO_FIXTURE_IDS.experienceTemplateId })
+    expect(referencedTemplate.sys.id).toBe(EXO_FIXTURE_IDS.experienceTemplateId)
+  })
+
+  // Both tests below run after the six above and re-import into the same space, exercising
```
ethan-ozelius-contentful, line 95: 🤔  It's worth double checking that these tests always run in the same order, and the this comment is actually true. Often times test suites run the `it` blocks in random order.

test/integration/import-exo.test.ts
```diff
@@ -0,0 +1,192 @@
...
+    const result = await runContentfulImport({
+      spaceId,
+      environmentId,
+      managementToken,
+      content: buildExoContent(EXO_FIXTURE_IDS),
+      includeExperienceOrchestration: true,
+      useVerboseRenderer: true
+    })
+
+    expect(result).toBeDefined()
```
ethan-ozelius-contentful, line 110: I'm assuming we don't need to check for anything additional here, correct?  If an error occurred, it would have been thrown on line 101: `await runContentfulImport({...`.  And nothing would change on a 2nd import, correct?


Update, I see more involved tests that make assertions on the content change in the next test.  Resolving this comment.

ethan-ozelius-contentful (approved):

test/integration/import-exo.test.ts
michaelphamcf: Nice call. Looked into it. Apparently, since jest v27 (we're now on `29`), each `it` block in a single file is executed in declaration order. Another way to randomize is to set the `--randomize` flag, but I don't see that being set anywhere. I think we're good.
https://jestjs.io/docs/29.7/setup-teardown#order-of-execution

PR #1696, opened 2026-08-13
fix: remove purpose and add taxonomy integration tests [AIS-134]

michaelphamcf:
[AIS-134](https://contentful.atlassian.net/browse/AIS-134)

## Summary
- Adds integration test coverage for ExO folder import. Covers: destination-scoped concept creation (prefLabel copied from a real source concept, and the fallback-to-derived-ID path when no source concept exists), linking new concepts into their parent folder-group scheme, in-place rewrite of entity `metadata.concepts`, re-import idempotency, and the same-space skip path.
- Removes dead purpose-patch logic in `lib/utils/import-exo-folders.ts` (per [review](https://github.com/contentful/contentful-import/pull/1696/changes#r3797933013)): the concepts API doesn't reliably return `purpose` in a GET response and doesn't allow updating it after creation, so the existing-concept patch path could never detect or repair it — only the destination space link is patched now. (Initially landed as a narrower guard rather than a full removal; confirmed empirically that *some* fix was load-bearing by reverting it and watching the `re-import is idempotent` test fail live as predicted, then removed the check entirely once review clarified it could never have worked at all.)
- Hardens the new tests' cleanup for concurrent CI: the cross-space `afterAll` now wraps org-scoped cleanup in try/finally so a cleanup failure can't leak the throwaway space, and shared parent-scheme patches retry on `VersionMismatch` since `test-integration` runs on every PR against the same org.
- Rebased onto latest `main` to pick up an unrelated fix ([15a17c1](https://github.com/contentful/contentful-import/commit/15a17c15eaed5347a3f74ba32a1000beeaa13b5d)).

## Test plan
- [x] `npm run test:unit` (248/248 pass)
- [x] `tsc --project tsconfig.test.json --noEmit`
- [x] Ran the integration suite live against the CI org via CI, confirming the fix is load-bearing and the rebase is green
- [ ] Address unrelated pre-existing lint errors reported by the repo-wide test lint command (out of scope for this PR)

[AIS-134]: https://contentful.atlassian.net/browse/AIS-134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/utils/import-exo-folders.ts
+++ b/lib/utils/import-exo-folders.ts
@@ -161,9 +161,7 @@ export async function createOrPatchChildConcepts(
     } else {
       const patches: Array<{ op: string; path: string; value: any }> = []
 
-      // The CMA does not include purpose in every concept GET response. Only
-      // patch it when the API actually returns a different purpose.
-      if (typeof (existing as any).purpose === 'string' && (existing as any).purpose !== 'internal') {
+      if ((existing as any).purpose !== 'internal') {
         patches.push({ op: 'add', path: '/purpose', value: 'internal' })
       }
 
--- a/lib/utils/import-exo-folders.ts
+++ b/lib/utils/import-exo-folders.ts
@@ -161,7 +161,9 @@ export async function createOrPatchChildConcepts(
     } else {
       const patches: Array<{ op: string; path: string; value: any }> = []
 
-      if ((existing as any).purpose !== 'internal') {
+      // The CMA does not include purpose in every concept GET response. Only
+      // patch it when the API actually returns a different purpose.
+      if (typeof (existing as any).purpose === 'string' && (existing as any).purpose !== 'internal') {
         patches.push({ op: 'add', path: '/purpose', value: 'internal' })
       }
 
--- a/lib/utils/import-exo-folders.ts
+++ b/lib/utils/import-exo-folders.ts
@@ -116,8 +116,7 @@ export function deriveChildConceptMap(
  * Step 3: Create or patch each destination child concept.
  * - If the concept doesn't exist: creates it with purpose:'internal', the source
  *   concept's prefLabel, and metadata.spaces scoped to the destination space.
- * - If it already exists: patches in any missing pieces (purpose:'internal' and/or
- *   the destination space link). Handles concepts created by earlier broken runs.
+ * - If it already exists: patches in the destination space link if missing.
  */
 export async function createOrPatchChildConcepts(
   plainClient: any,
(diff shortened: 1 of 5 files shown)


lib/utils/import-exo-folders.ts
```diff
@@ -161,7 +161,9 @@ export async function createOrPatchChildConcepts(
     } else {
       const patches: Array<{ op: string; path: string; value: any }> = []
 
-      if ((existing as any).purpose !== 'internal') {
+      // The CMA does not include purpose in every concept GET response. Only
```
ethan-ozelius-contentful: I wonder if we should remove the `purpose: internal` logic entirely 🤔 

From the team that owns concepts endpoint:

> There was a decision to omit purpose from API responses since it's not a public value. As an alternative for verifying whether the purpose=internal is there, you could try the concepts endpoint with purpose=internal as a filter
/organizations/:orgId/taxonomy/concepts?purpose=internal&sys.id[in]=:conceptId

...

> As pointed out, we don’t return purpose as you have to interact with these folders always via the filter-query parameter to be able to see them (for now). If you don’t provide this param always, you are interacting with normal concepts/schemes that show up in the Taxonomy Manager.

ethan-ozelius-contentful: Yeah, we should definitely remove this.

Comment from the referenced "decision":

>  i don't think we should allow it to be updated; only during creation.

So this patch wouldn't even work to add the purpose internal key/value.

lib/utils/import-exo-folders.ts
ethan-ozelius-contentful, line 140: Not related to this PR, but we can improve this typing with 

`let existing: ConceptProps | null = null` and import `ConceptProps` from CMA.js

ethan-ozelius-contentful (dismissed):

lib/utils/import-exo-folders.ts
michaelphamcf: This probably needs more than this one line change. Using `ConceptProps` causes an error on line 163 below. Holding off on this unless you think it should be changed in this PR.
```
Property 'metadata' does not exist on type 'LocalizedEntity<Omit<Concept, "conceptSchemes">, "prefLabel" | "altLabels" | "hiddenLabels" | "definition" | "editorialNote" | "historyNote" | "example" | "note" | "scopeNote", string>'.ts(2339)
```

michaelphamcf: Using `ConceptProps` causes an error on line 163 below.  This probably needs some changes in CMA to continue. Holding off on this unless you think it should be changed in this PR.
```
Property 'metadata' does not exist on type 'LocalizedEntity<Omit<Concept, "conceptSchemes">, "prefLabel" | "altLabels" | "hiddenLabels" | "definition" | "editorialNote" | "historyNote" | "example" | "note" | "scopeNote", string>'.ts(2339)
```

ethan-ozelius-contentful: Gotcha, makes sense.

ethan-ozelius-contentful (approved):

PR #1702, opened 2026-08-17
fix(publishing): respect per-locale publish state on import

Chaffexd:
## The bug

Contentful supports publishing individual locales of an entry, and `contentful-export` already serializes that state into `export.json` as `sys.fieldStatus`:

```json
"fieldStatus": { "*": { "en-US": "published", "es": "draft", "zh-Hant-TW": "changed" } }
```

`contentful-import` ignored it — before this change, `fieldStatus` appeared nowhere in `lib/`. In `lib/tasks/push-to-space/publishing.ts`, `runQueue()` did:

```ts
const publishedEntity = await requestQueue.add(() => entity.publish())
```

That's a whole-entry publish with no locale scope, and the CMA's defined behaviour is to mark **every** locale in the destination environment as published.

### Two symptoms

1. **All locales flip to Published on import** — including locales that contain no content at all, so the blast radius is wider than "entries with translations".
2. **Entries fail to publish entirely.** Where a draft locale lacks content for a required field, the whole-entry publish fails validation: `422 InvalidEntry: The property "es" is required here`.

### Reproduction

Reproduced end-to-end against a real space (144 entities: 138 entries + 6 assets carrying `fieldStatus`):

| Scenario | Result |
|---|---|
| Plain import | 129/144 entities with wrong per-locale state (only 7 had genuinely divergent locale content); 19 entries failed with `422 InvalidEntry` |
| Control: environment clone instead of export/import | State preserved perfectly — rules out anything environmental |
| Workaround: `--skip-content-publishing`, then replay `fieldStatus` via the CMA | 144/144 parity, 0 errors |

### Where this came from

A customer replicating Commerce Cloud SiteRestrictions into Contentful, modelling per-locale visibility by holding individual locales in Draft while the entry itself is Published. Export/import between spaces silently promoted every Draft locale to Published, destroying the model. The entities involved are ordinary entries (of a content type they call a Component), which is what this change covers — see "Scope" below.

## Why the fix belongs here

`lib/tasks/init-client.ts` builds the main client as `createClient(config, { type: 'legacy' })`. The legacy client's entity-level `publish()` is typed `(): Promise<Entry>` — no arguments, so the importer cannot express "publish only these locales".

The capability already exists in `contentful-management`, so **no SDK change is needed**:

- `locales?: string[]` on `entry.publish`/`unpublish` and `asset.publish`/`unpublish` (`plain-client-types.d.ts:225,229,273,277`)
- the REST adapter already builds the payload: `params.locales?.length ? { add: { fields: { '*': params.locales } } } : null`
- `sys.fieldStatus` is officially typed in `common-types.d.ts:235`

This repo already constructs a plain client — `initPlainClient` (`lib/tasks/init-client.ts:21`) is built unconditionally and threaded into `pushToSpace` as of the ExO work. This change consumes what's already there.

## The change

**1. The fix (default behaviour).** Publishing routes through the plain client's `entry.publish` / `asset.publish` with `locales` derived from `sys.fieldStatus['*']`, treating `published` and `changed` as live.

- `lib/utils/resolve-publish-locales.ts` (new) — pure resolver + plan builder
- `lib/tasks/push-to-space/publishing.ts` — `runQueue()` routes per entity
- `lib/tasks/push-to-space/push-to-space.ts` — builds the plan for assets and entries

**2. `unpublishDraftLocales` (new, opt-in, default off).** Publishing is additive — the payload verb is `add` — so it can add published locales but never demote one. This only bites when re-importing over already-published content. With the flag on, locales the content file marks `draft` that are still live in the destination are unpublished after publishing.

Demotions are computed by comparing the content file's `fieldStatus` against the **destination entity's own** `fieldStatus`, so no write is issued unless a locale is genuinely live there, and locales absent from the destination environment are never named. The unpublish uses the version from the publish response.

It is opt-in because it lets an import **unpublish** content in the destination — something the importer has never done.

### Behaviour is preserved wherever there's nothing to scope

The fallback is structural rather than something tests have to police. The resolver returns "publish as before" when:

- the content file has no `sys.fieldStatus` (older exports), **or**
- every destination locale is live anyway, so a whole-entity publish is equivalent, **or**
- the destination locales can't be read, **or**
- no plain client is available

If *none* of an entity's live locales exist in the destination, it is left unpublished with a warning rather than published across all locales.

### Where destination locales come from

`destinationData.locales` looks like the obvious source but isn't usable — it stays empty under `skipLocales`/`skipContentModel` (`get-destination-data.ts:266-275`), which would compute an empty intersection and skip publishing entirely. Locales are read from the environment instead, cached per environment, and **only fetched when at least one source entity actually carries `fieldStatus`** — so imports of older content files issue no extra API call at all.

## Backward compatibility

Existing export files already contain `sys.fieldStatus`, so **they benefit without being re-exported**. Content files that predate it take the unchanged code path.

## Scope

Covered: **entries and assets** — the only entities where `contentful-management` exposes a locale-scoped publish.

Not covered, for reference:

| Entity | `fieldStatus` in sys | Locale-scoped publish in SDK |
|---|---|---|
| Entry | yes | `locales: string[]` — **used here** |
| Asset | yes | `locales: string[]` — **used here** |
| ExO Experience | — | yes, but a different shape: `{ add: [...] }` payload |
| ExO Component | yes | none — `publish({ ..., version })` only |
| ExO Template / Fragment / DataAssembly | partly | none |

ExO Experiences could be supported later via their `{ add: [...] }` payload; ExO Components would need a `contentful-management` change first. Both are out of scope here.

## Remaining limitation

`changed` (published, with newer draft edits on top) is imported as `published`. Reconstructing it needs two separate writes and can't be expressed by an import. Documented in the README.

## On defaults: why the fix is default and the demotion is a flag

The core fix is default behaviour because it's a bugfix — the current behaviour destroys state the export file explicitly records, with no way to ask for it back short of the `--skip-content-publishing` + CMA-replay workaround. Happy to gate it if you disagree; it's the same seam.

The demotion is separate and opt-in because it changes an import from purely additive to something that can remove published content. The dominant workflow is export → import into a fresh environment, where it's a no-op; it earns its keep only on repeated imports.

**Flag naming is open** — `--unpublish-draft-locales` is descriptive but I'm not attached to it, and it's worth settling before release since it's permanent surface.

**Likely follow-up:** the flag is reachable via the module API immediately, but `contentful space import` goes through `contentful-cli`, which defines its own yargs options — so exposing the flag there is probably a separate PR worth confirming. The core fix needs no CLI change and reaches `contentful space import` users as soon as this releases.

## Test plan

```
Test Suites: 28 passed, 28 total
Tests:       287 passed, 287 total
```

Covering the four key cases plus edges:

- mixed per-locale state → only live locales published
- all-locales-published → unchanged whole-entity publish
- no `fieldStatus` → unchanged whole-entity publish, **and no locale lookup issued**
- locale absent from the destination environment → dropped from the publish
- `changed` treated as live; entity skipped when no live locale exists in the destination; fallback when locales can't be read; fallback when no plain client; assets routed to the asset endpoint; publish and demotion failures logged without aborting the import
- demotion: fires only for locales live in the destination, uses the post-publish version, no-ops when the entity is new or the flag is off
- `--unpublish-draft-locales` present in real CLI help output

Type check passes for `tsconfig.json` and `tsconfig.test.json`.

### Two things to be straight about

- **`npm run lint` exits 1 — but it does so identically on `main`** (104 pre-existing ESLint errors, mostly `comma-dangle` plus style in `test/unit/yargs.test.ts`). I diffed the rule-by-rule error list before and after: identical, 0 added. My files are clean. I deliberately left the pre-existing errors alone as unrelated churn — happy to fix them in a separate PR if wanted.
- **Integration tests were not run** — they need `MANAGEMENT_TOKEN` and `ORG_ID`, unavailable in my environment. CI will cover them.

### Validation status

The locale-scoped publish mechanism was verified live against a real space before this change was written: publishing with `locales: ['es']` moved only `es` and left the other locales untouched.

What has **not** been run is a full end-to-end import against a live space with this branch, nor the `--unpublish-draft-locales` demotion path. Those are covered by unit tests that pin the exact request shapes, but not yet by a live run. Flagging that explicitly rather than implying more validation than exists.


Chaffexd:
Superseded by #1704, which contains the same two commits pushed as a branch on this repository now that write access is sorted. Closing this one — please review #1704 instead.

PR #1704, opened 2026-08-18
fix(publishing): respect per-locale publish state on import

Chaffexd:
## The bug

Contentful supports publishing individual locales of an entry, and `contentful-export` already serializes that state into `export.json` as `sys.fieldStatus`:

```json
"fieldStatus": { "*": { "en-US": "published", "es": "draft", "zh-Hant-TW": "changed" } }
```

`contentful-import` ignored it — before this change, `fieldStatus` appeared nowhere in `lib/`. In `lib/tasks/push-to-space/publishing.ts`, `runQueue()` did:

```ts
const publishedEntity = await requestQueue.add(() => entity.publish())
```

That's a whole-entry publish with no locale scope, and the CMA's defined behaviour is to mark **every** locale in the destination environment as published.

### Two symptoms

1. **All locales flip to Published on import** — including locales that contain no content at all, so the blast radius is wider than "entries with translations".
2. **Entries fail to publish entirely.** Where a draft locale lacks content for a required field, the whole-entry publish fails validation: `422 InvalidEntry: The property "es" is required here`.

### Reproduction

Reproduced end-to-end against a real space (144 entities: 138 entries + 6 assets carrying `fieldStatus`):

| Scenario | Result |
|---|---|
| Plain import | 129/144 entities with wrong per-locale state (only 7 had genuinely divergent locale content); 19 entries failed with `422 InvalidEntry` |
| Control: environment clone instead of export/import | State preserved perfectly — rules out anything environmental |
| Workaround: `--skip-content-publishing`, then replay `fieldStatus` via the CMA | 144/144 parity, 0 errors |

### Where this came from

A customer replicating Commerce Cloud SiteRestrictions into Contentful, modelling per-locale visibility by holding individual locales in Draft while the entry itself is Published. Export/import between spaces silently promoted every Draft locale to Published, destroying the model. The entities involved are ordinary entries (of a content type they call a Component), which is what this change covers — see "Scope" below.

## Why the fix belongs here

`lib/tasks/init-client.ts` builds the main client as `createClient(config, { type: 'legacy' })`. The legacy client's entity-level `publish()` is typed `(): Promise<Entry>` — no arguments, so the importer cannot express "publish only these locales".

The capability already exists in `contentful-management`, so **no SDK change is needed**:

- `locales?: string[]` on `entry.publish`/`unpublish` and `asset.publish`/`unpublish` (`plain-client-types.d.ts:225,229,273,277`)
- the REST adapter already builds the payload: `params.locales?.length ? { add: { fields: { '*': params.locales } } } : null`
- `sys.fieldStatus` is officially typed in `common-types.d.ts:235`

This repo already constructs a plain client — `initPlainClient` (`lib/tasks/init-client.ts:21`) is built unconditionally and threaded into `pushToSpace` as of the ExO work. This change consumes what's already there.

## The change

**1. The fix (default behaviour).** Publishing routes through the plain client's `entry.publish` / `asset.publish` with `locales` derived from `sys.fieldStatus['*']`, treating `published` and `changed` as live.

- `lib/utils/resolve-publish-locales.ts` (new) — pure resolver + plan builder
- `lib/tasks/push-to-space/publishing.ts` — `runQueue()` routes per entity
- `lib/tasks/push-to-space/push-to-space.ts` — builds the plan for assets and entries

**2. `unpublishDraftLocales` (new, opt-in, default off).** Publishing is additive — the payload verb is `add` — so it can add published locales but never demote one. This only bites when re-importing over already-published content. With the flag on, locales the content file marks `draft` that are still live in the destination are unpublished after publishing.

Demotions are computed by comparing the content file's `fieldStatus` against the **destination entity's own** `fieldStatus`, so no write is issued unless a locale is genuinely live there, and locales absent from the destination environment are never named. The unpublish uses the version from the publish response.

It is opt-in because it lets an import **unpublish** content in the destination — something the importer has never done.

### Behaviour is preserved wherever there's nothing to scope

The fallback is structural rather than something tests have to police. The resolver returns "publish as before" when:

- the content file has no `sys.fieldStatus` (older exports), **or**
- every destination locale is live anyway, so a whole-entity publish is equivalent, **or**
- the destination locales can't be read, **or**
- no plain client is available

If *none* of an entity's live locales exist in the destination, it is left unpublished with a warning rather than published across all locales.

### Where destination locales come from

`destinationData.locales` looks like the obvious source but isn't usable — it stays empty under `skipLocales`/`skipContentModel` (`get-destination-data.ts:266-275`), which would compute an empty intersection and skip publishing entirely. Locales are read from the environment instead, cached per environment, and **only fetched when at least one source entity actually carries `fieldStatus`** — so imports of older content files issue no extra API call at all.

## Backward compatibility

Existing export files already contain `sys.fieldStatus`, so **they benefit without being re-exported**. Content files that predate it take the unchanged code path.

## Scope

Covered: **entries and assets** — the only entities where `contentful-management` exposes a locale-scoped publish.

Not covered, for reference:

| Entity | `fieldStatus` in sys | Locale-scoped publish in SDK |
|---|---|---|
| Entry | yes | `locales: string[]` — **used here** |
| Asset | yes | `locales: string[]` — **used here** |
| ExO Experience | — | yes, but a different shape: `{ add: [...] }` payload |
| ExO Component | yes | none — `publish({ ..., version })` only |
| ExO Template / Fragment / DataAssembly | partly | none |

ExO Experiences could be supported later via their `{ add: [...] }` payload; ExO Components would need a `contentful-management` change first. Both are out of scope here.

## Remaining limitation

`changed` (published, with newer draft edits on top) is imported as `published`. Reconstructing it needs two separate writes and can't be expressed by an import. Documented in the README.

## On defaults: why the fix is default and the demotion is a flag

The core fix is default behaviour because it's a bugfix — the current behaviour destroys state the export file explicitly records, with no way to ask for it back short of the `--skip-content-publishing` + CMA-replay workaround. Happy to gate it if you disagree; it's the same seam.

The demotion is separate and opt-in because it changes an import from purely additive to something that can remove published content. The dominant workflow is export → import into a fresh environment, where it's a no-op; it earns its keep only on repeated imports.

**Flag naming is open** — `--unpublish-draft-locales` is descriptive but I'm not attached to it, and it's worth settling before release since it's permanent surface.

**Likely follow-up:** the flag is reachable via the module API immediately, but `contentful space import` goes through `contentful-cli`, which defines its own yargs options — so exposing the flag there is probably a separate PR worth confirming. The core fix needs no CLI change and reaches `contentful space import` users as soon as this releases.

## Test plan

```
Test Suites: 28 passed, 28 total
Tests:       287 passed, 287 total
```

Covering the four key cases plus edges:

- mixed per-locale state → only live locales published
- all-locales-published → unchanged whole-entity publish
- no `fieldStatus` → unchanged whole-entity publish, **and no locale lookup issued**
- locale absent from the destination environment → dropped from the publish
- `changed` treated as live; entity skipped when no live locale exists in the destination; fallback when locales can't be read; fallback when no plain client; assets routed to the asset endpoint; publish and demotion failures logged without aborting the import
- demotion: fires only for locales live in the destination, uses the post-publish version, no-ops when the entity is new or the flag is off
- `--unpublish-draft-locales` present in real CLI help output

Type check passes for `tsconfig.json` and `tsconfig.test.json`.

### Two things to be straight about

- **`npm run lint` exits 1 — but it does so identically on `main`** (104 pre-existing ESLint errors, mostly `comma-dangle` plus style in `test/unit/yargs.test.ts`). I diffed the rule-by-rule error list before and after: identical, 0 added. My files are clean. I deliberately left the pre-existing errors alone as unrelated churn — happy to fix them in a separate PR if wanted.
- **Integration tests were not run** — they need `MANAGEMENT_TOKEN` and `ORG_ID`, unavailable in my environment. CI will cover them.

### Validation status

The locale-scoped publish mechanism was verified live against a real space before this change was written: publishing with `locales: ['es']` moved only `es` and left the other locales untouched.

What has **not** been run is a full end-to-end import against a live space with this branch, nor the `--unpublish-draft-locales` demotion path. Those are covered by unit tests that pin the exact request shapes, but not yet by a live run. Flagging that explicitly rather than implying more validation than exists.


lib/tasks/push-to-space/publishing.ts
```diff
@@ -73,13 +92,51 @@ export async function archiveEntities ({ entities, requestQueue }) {
...
+  if (!localesToDemote?.length) {
+    return published
+  }
+
+  logEmitter.emit('info', `Unpublishing locales ${localesToDemote.join(', ')} of ${entity.sys.type} ${getEntityName(entity)}`)
+
+  return plainClient[namespace].unpublish(
+    { spaceId, environmentId, [`${namespace}Id`]: entity.sys.id, locales: localesToDemote },
+    published
+  )
```
BobHemphill76: One thing I'd want to sort before the demotion path goes live: publish and unpublish are coupled into one unit here, so a failed demotion loses the successful publish. If the unpublish rejects, this function throws, so runQueue never pushes the entity into publishedEntities — which means line 155 classifies it as unpublished and line 164 recurses on it. The retry then re-sends the stale legacy entity.sys.version.

 <details><summary>Repro with a two-entity queue (one demotion failing)</summary>

 >>> publish attempts: [ 'entry-ok@v7', 'entry-demote-fails@v7', 'entry-demote-fails@v7' ]
 >>> result ids: [ 'entry-ok' ]
 >>> ERROR: unpublish entry-demote-fails failed
 >>> ERROR: 409 VersionMismatch: sent 7, server is at 8
 >>> ERROR: Could not publish the following entities: entry-demote-fails
 </details>

 So the operator gets three errors where there's really one, and the last line is actively wrong — the entity was published, only the demotion failed. That's the part that worries me more than the redundant write: it points triage at the wrong entity and the wrong failure.

Fix is small — let the demotion fail on its own:

```suggestion
 try {
     return await plainClient[namespace].unpublish(
        { spaceId, environmentId, [`${namespace}Id`]: entity.sys.id, locales: localesToDemote },
         published
     )
     } catch (err: any) {
         logEmitter.emit('error', err)
        return published   // publish succeeded — don't let runQueue retry it
}
```

lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -776,15 +796,88 @@ function archiveEntities({ entities, sourceEntities, requestQueue }) {
...
+/**
+ * Locale codes of the destination environment, or null when they cannot be read.
+ * `destinationData.locales` is not usable here — it stays empty under
+ * `skipLocales`/`skipContentModel`.
+ */
+function getDestinationLocaleCodes(environment, requestQueue): Promise<string[] | null> {
+  let localeCodes = destinationLocaleCodesCache.get(environment)
+
+  if (!localeCodes) {
+    localeCodes = requestQueue.add(() => environment.getLocales())
```
BobHemphill76: Curious whether we want to page this one.  The repo already pages this exact collection. get-destination-data.ts:14 registers locales: { method: 'getLocales' } and batchedPageQuery (lines 72-105) walks it with limit: 100 until total is exhausted. This call passes no limit or skip at all and I believe we've had bugs in the past where we fail to fetch more than 100 locales (the default limit)

BobHemphill76 (dismissed):

lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -776,15 +797,90 @@ function archiveEntities({ entities, sourceEntities, requestQueue }) {
   return publishing.archiveEntities({ entities: entitiesToArchive, requestQueue })
 }
 
-function publishEntities({ entities, sourceEntities, requestQueue }) {
+type LocalePublishingSetup = {
+  plainClient: any
```
ethan-ozelius-contentful: let's add types: `plainClient: PlainClientAPI` & `environment: EnvironmentProps`

lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -776,15 +797,90 @@ function archiveEntities({ entities, sourceEntities, requestQueue }) {
...
+  /** Destination entities, passed only when `unpublishDraftLocales` is enabled. */
+  destinationEntitiesById?: Map<string, any>
+}
+
+// Cached per environment so assets and entries share one lookup per import.
+const destinationLocaleCodesCache = new WeakMap<object, Promise<string[] | null>>()
+
+/**
+ * Locale codes of the destination environment, or null when they cannot be read.
+ * `destinationData.locales` is not usable here — it stays empty under
```
ethan-ozelius-contentful: Can I get some more info on why _`destinationData.locales` is not usable here`_?

lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -776,15 +797,90 @@ function archiveEntities({ entities, sourceEntities, requestQueue }) {
...
+    destinationLocaleCodesCache.set(environment, localeCodes as Promise<string[] | null>)
+  }
+
+  return localeCodes as Promise<string[] | null>
+}
+
+async function publishEntities({ entities, sourceEntities, requestQueue, localePublishing }: {
+  entities: any[]
+  sourceEntities: any[]
+  requestQueue: any
```
ethan-ozelius-contentful: We can use `PQueue` type here for requestQueue.  I see that the code did not previously type `entities`, but if all we're publishing is entries/assets we can use `EntryProps/AssetProps`

lib/tasks/push-to-space/push-to-space.ts
ethan-ozelius-contentful: I'm a little confused by this, and why we can't re-use the existing [fetch locales code](https://github.com/contentful/contentful-import/blob/migrate-cma-plain-client/lib/tasks/get-destination-data.ts#L289-L301)

ethan-ozelius-contentful:
Couple things here:

1. Locale based publishing is an entitlement that not all customers have, so we should have some entitlement checking/error handling for spaces that lack the entitlement.
2. I have an existing [PR](https://github.com/contentful/contentful-import/pull/1670) to migrate to use plain client exclusively that should be merged before this PR, which might make this implementation more simple.
3. I'm a little nervous about mixing/matching the content file and destination data in `resolveLocalesToDemote()`, I think we should rely on the content file as the source of truth.
4. I wonder if once this [PR](https://github.com/contentful/contentful-import/pull/1670) is merged the implementation could be just:
```ts
- return client.entry.publish({ spaceId, environmentId, entryId: id }, entity)
+ return client.entry.publish({ spaceId, environmentId, entryId: id, locales }
```

lib/tasks/push-to-space/push-to-space.ts
Chaffexd: Done in d380f69 — with one deliberate deviation: the field stays `Environment` (the SDK entity) rather than `EnvironmentProps`, because `batchedPageQuery` calls `environment.getLocales()` and `EnvironmentProps` is a plain object with no methods. There's a comment on the field saying so.

Typing `plainClient` as `PlainClientAPI` also flushed out why it had been `any`: `plainClient[namespace].publish({ ..., [`${namespace}Id`]: id })` can't satisfy the typed `entryId`/`assetId` params through a computed key. Split into `localeScopedPublish`/`localeScopedUnpublish` which branch on `namespace`, so both calls are genuinely typed now.

That caught a latent bug on the way — the `unpublish` call was `(localePublishing, entity, published)`, i.e. `published` sitting in the `locales` slot. It was invisible because `published` was implicitly `any`.

lib/tasks/push-to-space/push-to-space.ts
Chaffexd: We do reuse it — `getDestinationLocaleCodes` calls the exported `batchedPageQuery` with `type: 'locales'`, so the paging loop and the `getLocales` registration in `OFFSET_QUERY_METHODS` are shared. Nothing is reimplemented.

What isn't reusable is the *result field*, `destinationData.locales`. It's only populated when `!skipContentModel && !skipLocales` **and** the content file itself carries locales (get-destination-data.ts:266-274); otherwise it stays `[]`. And `[]` is indistinguishable from "the destination has no locales" — which would make `resolvePublishLocales` return an empty array for every entity and skip the entire publish.

Publish scoping needs the real locale list whatever the skip flags say, so it asks for them directly and caches the answer per environment for the run (entries and assets share one lookup). I've expanded the docstring to spell this out rather than leaving the one-liner that prompted the question.

lib/tasks/push-to-space/push-to-space.ts
Chaffexd: `requestQueue: PQueue` done in d380f69.

On the entities, I went with `(Entry | Asset | ContentType)[]` rather than `EntryProps/AssetProps`, for two reasons:

- Content types go through this function too — see the "Publishing Content Types" task at push-to-space.ts:219, which calls the same `publishEntities`.
- They're SDK entities, not `*Props`. `publishing.ts` calls `entity.publish()` on them, which only exists on the entity. (This is the part that collapses once #1670 is in — see my reply on the main thread.)

Chaffexd:
Thanks — worked through all four.

**1. Entitlement checking.** Agreed, and fixed in d380f69. A 403 on the locale-scoped publish now emits one warning and falls back to a whole-entity publish, so an unentitled space imports exactly as it did before this feature instead of failing. Tracked in a `WeakSet` keyed on the client, so the entry and asset passes share a single warning and subsequent entities skip the doomed attempt (including demotions, which would 403 too). A genuine permissions 403 isn't masked — the fallback publish then fails on its own and reports normally. Three tests cover it, plus a README note.

One thing worth your eyes: I keyed the check on HTTP status because I don't know the actual response shape for a missing locale-publishing entitlement. If it comes back with a `details.reasons` string like `exoM1 entitlement required` does, I'd rather match that precisely — do you know?

**2 + 4. #1670.** This is already merged (53e7b9d), which is what's making this PR show conflicts — `publishing.ts`, `push-to-space.ts` and `get-destination-data.ts`. So I need a merge from main here regardless, and you were right that it simplifies things. Concretely, on top of main's `publishEntity(client, spaceId, environmentId, entity)`:

- `localeScopedPublish`/`localeScopedUnpublish` collapse into passing `locales` through to the existing `client.entry.publish(...)` call — your point 4 exactly.
- The `LocalePublishing` plumbing type, its `namespace` field and the legacy-vs-plain branch in `publishEntityLocales` all go away; `runQueue` already carries `client`/`spaceId`/`environmentId`.
- The `Environment`-vs-`EnvironmentProps` question on the other thread probably dissolves too, depending on where `batchedPageQuery` ended up.

What survives either way: `resolve-publish-locales.ts` (the plan), the destination locale fetch, and the entitlement fallback above.

I'd rather do that merge as its own commit so the simplification is reviewable separately from the behaviour, rather than smuggling it into a conflict resolution. Happy to push it next if you agree.

**3. Source of truth in `resolveLocalesToDemote`.** I'd push back gently here, and defer to you. The content file *is* the source of truth: only locales it marks `draft` are ever candidates for demotion. The two destination reads aren't inputs to that decision — they're write-avoidance filters:

- `destinationLocaleCodes` — never name a locale the destination doesn't have.
- `destinationFieldStatus` — only unpublish where the locale is actually live there today.

Drop them and we issue an unpublish for every draft locale of every entity, including ones that were never published in the destination. That's N wasted writes per import at best, and possibly 400s. But if you'd rather have fewer moving parts and eat the extra writes, say so and I'll cut them — it's a small change and `unpublishDraftLocales` is opt-in anyway.

ethan-ozelius-contentful:
Hello, I'm still wrapping my head around this code and all the discussion, and I might be missing something, but I'm not crazy about this "demotion" model.  I think it overly complicates something that might be achievable in just a few lines of code.

Would this achieve the goal for the customer's use case?

```ts
function publishEntity(client: PlainClientAPI, spaceId: string, environmentId: string, entity: any): Promise<any> {
  const id = entity.sys.id
  const type = entity.sys.type
  + const allLocales = entity.sys.fieldStatus && entity.sys.fieldStatus["*"] || {}
  + for (const [key, value] of Object.entries(allLocales)) {
  +  if (value === 'published') {
  +     publishedLocales.push({
  +      [key]: value
  +      })
  +  } 
  + }

  if (type === 'Entry') {
    - return client.entry.publish({ spaceId, environmentId, entryId: id }, entity)
    + return client.entry.publish({ spaceId, environmentId, entryId: id, locales: publishedLocales }, entity) 
...
// same for asset and content type.
```

ethan-ozelius-contentful:
Hello @Chaffexd @matthewheath, 

Few things:
1. merge conflicts are still present preventing the CI checks from running.
2. I checked this code out and ran a test, there's an issue when locale based publishing is not enabled for the destination space.
- setup a space with locale based publishing (en-US, de & es), create an entry with a published en-US, draft de, and changed es status.  
- take an export, and attempt to import into a new space that does not have locale based publishing enabled.
- observe that the entry has only one status of published.

We'll need to add an additional step to ensure that locale based publishing is enabled before adding this functionality.

3. I still think we should explore [this option](https://github.com/contentful/contentful-import/pull/1704#issuecomment-5441784836) some more, the implementation would be **much** cleaner.  I'm really not thrilled about using a "demotion" strategy when it should be possible to use the existing publish mechanism, just passing an additional param that is already supported.

lib/tasks/push-to-space/publishing.ts
```diff
@@ -73,13 +93,119 @@ export async function archiveEntities ({ entities, requestQueue }) {
   return allArchivedEntities
 }
 
-async function runQueue (queue, result: ResourcesUnion = [], requestQueue: PQueue) {
+/**
```
ethan-ozelius-contentful, line 96: All these helpers should live in their own util files.

lib/utils/resolve-publish-locales.ts
```diff
@@ -0,0 +1,141 @@
...
+ * `contentful-export` serializes per-locale publish state as `sys.fieldStatus`:
+ *
+ *   "fieldStatus": { "*": { "en-US": "published", "es": "draft", "zh-Hant-TW": "changed" } }
+ *
+ * A whole-entry publish has no locale scope, so the CMA marks every locale in the
+ * destination environment as published. These helpers work out which locales a
+ * publish should be scoped to so that exported state survives the import.
+ */
+
+const LIVE_FIELD_STATUSES = ['published', 'changed']
```
ethan-ozelius-contentful, line 11: This would still cause `"changed"` status entities to be published, I'm not sure we want that.

lib/tasks/push-to-space/publishing.ts
Chaffexd: Done. `publishing.ts` is orchestration only now — `publishEntity`, `unpublishEntityLocales`, `archiveEntity`, `publishEntityLocales`, `runQueue`.

The locale logic moved into three utils:

- `lib/utils/resolve-publish-locales.ts` — builds the per-entity publish plan from the source `sys.fieldStatus`
- `lib/utils/destination-locales.ts` — reads and caches the destination environment's locale codes
- `lib/utils/locale-publishing.ts` — the entitlement check, the "did the destination honour the scope" check, and the one-warning-per-run latch

Chaffexd:
@ethan-ozelius-contentful @matthewheath — all three are addressed and pushed. CI is green.

#1670 landing made a real difference here: with the plain client threaded through `publishing.ts`, most of the machinery behind point 3 just evaporated.

## 1. Merge conflicts

Resolved against `main`. Checks are running again and all pass — `build`, `check / lint`, `test-integration`, CodeQL, Wiz.

## 2. Entitlement

Reproduced your case. There are **two** independent gates, which is why the 403 fallback alone didn't cover it:

| Gate | Scope | How we detect it |
|---|---|---|
| Locale-based publishing entitlement | organization / plan | `GET /organizations/{orgId}/organization_entitlement_set`, checked up front |
| `Locale-based (un)publishing` selected under **Settings → Locales → Publishing options** (default is "Publish all locales") | **per environment** | the publish response |

Your repro is the second one — a brand-new space in an otherwise entitled org. The CMA does not reject the `{ add: { fields: { '*': [...] } } }` payload there; it accepts it and publishes the whole entity anyway. That is exactly the "only one status of published" you saw, and it is why nothing 403'd.

So `lib/utils/locale-publishing.ts` now does both:

- **Entitlement**, up front: `spaceHasLocalePublishingEntitlement` mirrors the existing `spaceHasExoM1Entitlement` exactly — same public endpoint, same `features[...].value` shape, same "`null` means inconclusive, keep going" contract. Not entitled → we never scope, one warning, whole-entity publish, and we don't even bother reading the destination locales.
- **Environment setting**, from the first scoped publish: `sys.fieldStatus` is the destination's own per-locale state, and an environment that doesn't track per-locale state doesn't return it. No `fieldStatus` in the response ⇒ the scope was ignored ⇒ latch off for the rest of the run, warn once naming the setting, and stop attempting demotions (they'd be pointless writes producing confusing version-mismatch errors).

`400` is now treated as a rejection alongside `403`. Both gates funnel into one latch and one warning, and neither fails the import — worst case is exactly the pre-feature behaviour.

**One thing I could not verify myself:** whether an environment on "Publish all locales" *ever* returns `sys.fieldStatus`. I have no space with the setting off to test against. You've already got both spaces set up — could you re-run your repro and confirm you now get the warning rather than silence? If it turns out `fieldStatus` comes back regardless, I'll swap that signal for a comparison against the requested locales instead. The design fails safe either way: if the signal is wrong we lose the feature loudly, we never write wrong state.

## 3. Simplification

`publishEntity` is now what you sketched:

```ts
function publishEntity(client, spaceId, environmentId, entity, locales?: string[]) {
  const id = entity.sys.id
  const type = entity.sys.type
  if (type === 'Entry') {
    return client.entry.publish({ spaceId, environmentId, entryId: id, locales }, entity)
  }
  if (type === 'Asset') {
    return client.asset.publish({ spaceId, environmentId, assetId: id, locales }, entity)
  }
  if (type === 'ContentType') {
    // Content types are not localized, and contentType.publish takes no locales.
    return client.contentType.publish({ spaceId, environmentId, contentTypeId: id }, entity)
  }
  ...
}
```

Gone: the `LocalePublishing` client-context object, the entry/asset namespace branching, the `localeScopedPublish`/`localeScopedUnpublish` wrappers, and the legacy `entity.publish()` fallback. That's ~50 lines out of `publishing.ts`.

The one part of the sketch I didn't take is reading `entity.sys.fieldStatus` at the publish site — see below.

## On the agent comment

It holds up. I checked the three load-bearing claims against the code rather than taking them on trust:

- `getPlainData` is `omit(entity, 'sys')` (`creation.ts:296`), so the source `sys` never goes over the wire. The `sys.fieldStatus` on a CMA response is the **destination's** state, not the source's.
- `createEntries` short-circuits with `return entry.transformed` when `skipUpdates` hits an existing entity (`creation.ts:130`). So `entity.sys` at the publish site would carry source state on that one path and destination state on every other one — reading it inline really would make the behaviour depend on `skipContentUpdates`.
- `contentType.publish` sends `null` as the body and takes no `locales`, and the SDK gates on `params.locales?.length` — so an empty list silently degrades to an unscoped publish.

That's why the plan is still built from `sourceEntities[].original.sys` in `resolve-publish-locales.ts`: it's the only place the source `fieldStatus` still exists at that stage.

Where the agent comment is now **out of date**: it argued that collapsing to `publish({ locales })` would mean plumbing `sourceEntities` down and "would not be smaller". Post-#1670 that's no longer true — the plumbing was already there, and the collapse made it meaningfully smaller. So the simplification you asked for is in; only the *source* of the locale list differs from your sketch, and that part the agent got right.

## Tests

52 new/updated tests covering locale publishing, including your repro as a unit test (`detects a destination that accepts the payload but ignores the scope`) and the entitlement gate in all four states — entitled, not entitled, feature key absent, check failed. `tsc --noEmit` clean.

michaelpineirocontentful (approved):

chasepoirier (approved):

PR #1709, opened 2026-09-08
feat: import Optimization Variants for Experiences and Fragments [AIS-139]

michaelphamcf:
[AIS-139](https://contentful.atlassian.net/browse/AIS-139)

## Summary
- Adds import support for Experience and Experience Fragment Optimization Variants, reading the nested `experience.optimizationVariants` / `experienceFragment.optimizationVariants` shape produced by the paired `contentful-export` change ([contentful/contentful-export#2375](https://github.com/contentful/contentful-export/pull/2375)).
- Variants don't preserve their source ID on import. Every other ExO entity type is created with the same ID it had in the source space, but the upstream API always generates a brand-new ID for a variant on create and has no way to create one with a specific ID. So a variant that already exists in the destination space, having been created there by a previous import, will get created again as a duplicate rather than updated. This is called out explicitly in the README and docs.
- Variant creation and publishing go through new, self-contained functions rather than reusing the existing per-entity helpers — those helpers assume an entity's ID is unique within its type, which isn't true for variants (a variant's ID is borrowed from its parent).
- Filters out the same API-synthesized "default" entry that `contentful-export` already excludes, as a defensive backstop for older export files.
- Also included: live integration test coverage against a real space, README/doc updates (including two import-order steps that were missing from the existing list), and a dependency swap from a local workaround package to the real published `contentful-management@^12.17.0`.

## Test plan
- [x] Unit suite passes (279/279) — one pre-existing, unrelated test-suite failure on `main` predates this change (confirmed via `git stash`)
- [x] New integration tests run live against a real space — confirm variant creation, publish-state carrying over correctly (one published variant, one draft), and correct parent references
- [x] Manually verified a full export → import round trip against a real sandbox space, reading back the destination directly rather than trusting the import tool's own log — correct variant counts and names, no duplicates

[AIS-139]: https://contentful.atlassian.net/browse/AIS-139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -709,6 +709,87 @@ export default function pushToSpace({
...
+          parents: ctx.data.experiences || [],
+          sourceParents: sourceData.experiences || [],
+          client,
+          spaceId,
+          environmentId,
+          kind: 'experience'
+        })
+        ctx.data.experienceVariants = results.flatMap((r) => r.created)
+      }),
+      skip: () => !includeExperienceOrchestration || !(sourceData.experiences || []).some((e: any) => e.optimizationVariants?.length)
```
whitelisab: The skip conditions here don't rely on a flag, only data presence. Is there a chance that re-running an import against the same destination will create duplicate variants? Not sure if this is something that we'd want to consider, but perhaps adding a `skipExoVariants` flag (mirroring the flag in export) or a runtime warning logged when the task fires might be helpful?

lib/tasks/push-to-space/push-to-space.ts
```diff
@@ -788,6 +869,68 @@ function omitSys(entity) {
...
+}): Promise<Array<{ parentId: string; created: any[] }>> {
+  const ns = kind === 'experience' ? client.experienceVariant : client.experienceFragmentVariant
+  const idParam = kind === 'experience' ? 'experienceId' : 'experienceFragmentId'
+  const label = kind === 'experience' ? 'ExperienceVariant' : 'ExperienceFragmentVariant'
+  // Create requires the parent-reference field back (mirrors why the parent's own create call
+  // re-adds `entity.sys.experienceTemplate`/`entity.sys.component` after omitSys elsewhere in this file).
+  const parentRefField = kind === 'experience' ? 'experienceTemplate' : 'component'
+
+  return Promise.all(parents.map(async (parent) => {
+    const sourceParent = sourceParents.find((p) => p.sys.id === parent.sys.id)
```
whitelisab: This line is O(n²) in parent count. Probably not an issue at this scale, but worth cleaning up if it ever needs to scale.

whitelisab (approved):
Looks good, added some small comments for you to consider.

lib/tasks/push-to-space/push-to-space.ts
michaelphamcf: Great suggestion! Since the API generates a new variant ID on every create, re-running the import can create duplicates. I added an optional `skipExoVariants` flag, defaulting to false to preserve the current behavior, plus a runtime warning when variants are imported.

lib/tasks/push-to-space/push-to-space.ts
michaelphamcf: Thanks for pointing this out. Replaced the per-parent sourceParents.find(...) lookup with a precomputed map keyed by parent ID. Should be improved to something like O(n) now.

whitelisab (approved):

PR #1711, opened 2026-09-09
feat: support experience releases import [AIS-140]

michaelphamcf:
[AIS-140](https://contentful.atlassian.net/browse/AIS-140)

## Summary
- Adds Releases (Timeline) import, reading the `releases` array produced by the paired `contentful-export` change ([contentful/contentful-export#2376](https://github.com/contentful/contentful-export/pull/2376)). Only `Release.v2` ("Releases") is imported — any `Release.v1` ("Launch") release is skipped and logged as an error, since its payload shape isn't compatible with the `Release.v2` API.
- A release already existing in the destination space at the source's exact `sys.id` is updated; otherwise one is created. **Releases have no ID-preserving create, unlike every other importable entity** — the API always server-generates a release's `sys.id`, so re-importing the same source data creates additional releases rather than updating the ones from a prior run. This is called out explicitly in the README.
- Requires the destination organization to have the Releases entitlement (a separate, paid feature, gated independently from Experience Orchestration's `exoM1` entitlement) — a destination without it fails that release's import with a clear error rather than aborting the whole run.
- The Releases fetch/import logic lives in its own file (`lib/tasks/push-to-space/releases.ts`), matching how `assets`/`creation`/`publishing` are already organized in this codebase, with direct unit test coverage for each function.

## Test plan
- [x] Unit suite passes (284/284)
- [x] `tsc --noEmit` and lint clean
- [x] New integration test run live against a real sandbox space — imports a release referencing a real entry, confirms it's created at a server-generated ID (not the source's), with title/schemaVersion/entities intact
- [x] Manually verified a full export → import round trip against a real sandbox space, paired with the corresponding `contentful-export` branch

[AIS-140]: https://contentful.atlassian.net/browse/AIS-140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/transform/transform-space.ts
```diff
@@ -1,23 +1,26 @@
 import { omit, defaults } from 'lodash-es'
```
whitelisab: I think there's an unused import here (defaults)

lib/tasks/get-destination-data.ts
```diff
@@ -32,6 +33,7 @@ const ENTITY_METHODS = {
   assets: { name: 'assets', ns: 'asset' },
   locales: { name: 'locales', ns: 'locale' },
   tags: { name: 'tags', ns: 'tag' },
+  releases: { name: 'releases', ns: 'release' }
```
whitelisab: I think this releases entry is dead code. `ENTITY_METHODS` is only read by `batchedIdQuery` and `batchedPageQuery` (lines 64–90), and neither is ever called with type: 'releases'; the actual Releases fetch uses `CURSOR_QUERY_METHODS` above. Recommend deleting this line so the map doesn't imply a code path that doesn't exist.

lib/transform/transform-space.ts
```diff
@@ -1,23 +1,26 @@
...
 import sortEntries from '../utils/sort-entries'
 import sortLocales from '../utils/sort-locales'
 import { upgradeExoResources } from './exo-rename'
 import { DestinationData, OriginalSourceData, TransformedSourceData } from '../types'
 
-const spaceEntities = [
-  'contentTypes', 'entries', 'assets', 'locales', 'webhooks', 'tags'
+// POSSIBLE TODO: should experience orchestration entities be included here? If so, we need to update the type accordingly.
+// exo entities are added to this default functions return entities via `upgradeExoResources()`
+// Ethan's take Sept 4th 2026, functionally it doesn't make a difference because exo entities DO get added to the returned
```
whitelisab: Maybe a bit too specific of a detail to include in comments?

whitelisab (commented):
I noted a couple small things to clean up. There's also several TODOs in here which is totally fine, but just checking that you have a path to get those TODOs addressed later.

lib/transform/transform-space.ts
michaelphamcf: removed

lib/tasks/get-destination-data.ts
michaelphamcf: indeed dead code. removed

lib/transform/transform-space.ts
michaelphamcf: Went into a little bit of a rabbit hole with this one and created a ticket to follow up. Edited the comment.

I also edited the other TODOs, adding some more info, referencing a ticket that has more details.

whitelisab (approved):
🚀

